PR #30, opened 2013-07-23
Fix last modified field content for resources in dbbrowser

ljo:
And obviously some of the old not merged stuff too.


dizzzz:
looks fine, though.... not sure about the '1'  still ; would a space or something else work? Maybe a xqdoc header with a creation date + username ?

PR #33, opened 2013-08-20
redirect /exist/apps/eXide to /exist/apps/eXide/

joewiz:
without the trailing slash, the browser would prompt users to authenticate, leading to confusion.


dizzzz:
@wolfgangmm the change looks OK or......... are there strings attached?

PR #35, opened 2013-09-04
Add public getEditor method to retrieve private editor

wshager:
Convenience method for outsiders.


wolfgangmm:
Merged into develop manually.

PR #51, opened 2014-02-14
Update eXide.min.js

jensopetersen:
Queries do not only find, but construct, so "returned" is a better word than "found". The popup will have to be made a little wider. Perhaps the "s" on "item" should be made conditional on "+h+".


jensopetersen:
Yes, but the non-minified script is not in the repo.

joewiz:
The pre-minified source is in src/eXide.js. Modifying that will ensure your change persists.

jensopetersen:
Sorry - didn't notice that. Will modify.

PR #52, opened 2014-02-14
Update keybindings.js

jensopetersen:
This (handy) key combo appears vacant. Because F12 is reached on a MacBook using two keys, fn and F12, the present key combo for toggleQueryResults requires four keys, which is rather unpractical.


dizzzz:
The fn-F12 key combo is a choice on your mac; it can be changed into one.....

jensopetersen:
Yes, but E is vacant, so why use F12? E can be used by all.

PR #59, opened 2014-05-30
Allow saving of empty binary docs

wshager:


wshager:
The check is not ok, it obviously fails when the data is binary.

PR #71, opened 2014-09-10
Directory

christophe-g:
# new feature : sublime-style collection/resource tree editor

I like the way sublime-text allows to browse resources directly in the editor, so thought about cloning this feature. This is a rapid first go. 

The pull request splits the outline editor into two : the classical `outline` + a new area for browsing database collections and resources called `directory`:

exide-directory

It is pretty straightforward to use it: click on collections to open/close them, click on resources to open them in the editor. 

Some thought for future development: 
- add drag and drop + rename features 
- find good way to display/modify resources permissions
- keep tree state for later sessions (and also when re-opening a collection)

Also added in the pull, some small fix to make the git features work again with version 2.1.0.

Cheers,
Christophe

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="UTF-8"?>
 <project basedir="." default="all" name="eXide">
     <property file="local.build.properties"/>
     <property file="build.properties"/>
@@ -67,6 +66,7 @@
                 <file name="javascript-helper.js"/>
                 <file name="css-helper.js"/>
                 <file name="outline.js"/>
+                <file name="directory.js"/>
                 <file name="debuger.js"/>
                 <file name="codevalidator.js"/>
                 <file name="editor.js"/>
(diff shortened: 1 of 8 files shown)


joewiz:
Whoo hoo!  I love it already from the screenshot and can't wait to give it a try.  Is there a way to force the view to refresh, since presumably collections and documents could change after the view is first created?

wolfgangmm:
Looks great! A few minor things to address:
- as Joe already suspected, the view does not update if you store a new file into a collection. I guess the directory browser should listen for update events (which we may have to create for the purpose).
- when loading eXide, none of the two tabs is shown as active
- there should be some protection against loading very large collections. My suggestion would be to call collection.xql with a limit (in the end parameter) and only show the first 100 docs with a note that the collection is too large to be displayed entirely.

christophe-g:
Thanks for the comments !
- re refresh, you are right, the view is not updated automatically. For the moment, closing and reopening a collection will update the view (as I do not use any notion of tree state, closing collections removes all child nodes and expanding them again triggers a call to the database to refresh the view). 
- problem with on loading exide (tabs not active): I'll commit an update tomorrow. That should be an easy fix. 
- good point re very large collections. It would be worth testing very large collections to see at which point rendering starts to be a problem.

Cheers, 
C.

PR #72, opened 2014-09-11
outline tab activated on load

christophe-g:
make the outline tab look active when we open eXide.


christophe-g:
It is pretty handy to browse the db manager with the directory tree!

This pull request will add a sync feature between the directory view and the db manager, only when it is visible (from the view to the db manager, so it's a one way sync for the moment). 

Cheers,
C.

PR #79, opened 2014-11-12
Fix problem on state restore

christophe-g:
Got eXide stall on state restore (do not have time to investigate further...).
This fixes it.
Cheers,
C.


christophe-g:
Well I don't know what happened here, but the browser was receiving index.html without being processed through view.xql, hence missing the following bit.  

```
<script type="text/javascript">
                    eXide.namespace("eXide.configuration");
                    eXide.configuration.allowExecution = true;
                    eXide.configuration.allowGuest = true;
                    eXide.configuration.context = "";
                </script>
```

Strange ... 
You can ignore this pull request though - the real problem lies somewhere above.
C.

PR #81, opened 2014-11-23
Fix outline being called too often

christophe-g:
Outline can be pretty resource intensive in some case (e.g. with complex import tree) 
This pull request makes sure outline is called only when the outline view is active. With current behavior, outline view also is being updated when directory view is active. 
Cheers,
C.

### DIFF
--- a/resources/css/eXide.css
+++ b/resources/css/eXide.css
@@ -1477,7 +1477,6 @@ div#valueHighLight {
 .tree li {
     list-style-type:none;
     margin:0;
-    display:table;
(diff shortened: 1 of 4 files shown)


christophe-g:
Well it is a bit better, but even with those commits,`outline.xql` is called way too often (and is making cpu surge drastically in some cases). We might need a timeout mechanism here. 
I have the impression the outline was being updated only when a tab was activated, and not after every code modification, as it seems to be the case now. 
Cheers, 
C.

PR #88, opened 2015-02-12
Make checkbox labels clickable

joewiz:


joewiz:
one more file to commit - should be coming shortly

joewiz:
Okay, that's the full request. Sorry for the unintended "merge branch" commit. Happy to close and resubmit if you'd prefer a clean pull request.

joewiz:
This is the result of the change:

screen shot 2015-02-12 at 10 45 28 am

Now the labels, e.g., execute, can be clicked directly, rather than having to target the checkbox itself.

PR #89, opened 2015-02-22
Git

christophe-g:
Allowing the sync directory to be different from the git directory. Previously it was assumed that app is the root of the git repo. 

Previous Merge conflict resolved. 

file `config.xml` to be placed on the app root collection now reads : 

``` xml
<git xmlns="http://exist-db.org/eXide/git">
    <syncDir>/home/yourUserName/Directory for syncing the app</syncDir>
    <gitDir>/home/yourUserName/Directory where git resides</gitDir>
</git>
```

note change of name from `workingDir` to `syncDir` 
It was previously:

``` xml
<git xmlns="http://exist-db.org/eXide/git">
    <workingDir>/home/yourUserName/workingDir</workingDir>

</git>
```

Cheers
C.


joewiz:
@wolfgangmm This would be a clean merge, but could you take a look when you have the chance? My only concern is that the git configuration format changes, and I bet this change would disrupt someone who upgrades (there is no warning or upgrade assistance). Perhaps some "migration assistant" should be added? 

@christophe-g Your thoughts? Could you provide some explanation about why one might want to distinguish between the directory in source control and the sync directory?

christophe-g:
@joewiz This PR was an answer to a message from T. Cook who had his sync directory part of a larger repository - hence the need for more flexibility on the setting. 
Re upgrade disruption - I am not sure the git feature is widely used - or known ; )

joewiz:
@twcook Are you still interested in the functionality provided in this pull request? Otherwise we will likely close the request.

joewiz:
@christophe-g Thanks for the context. Without a migration assistant in place, I am inclined to close the PR. Unless you are invested in this functionality and would like to add in something to help migrate people who used the original settings?

joewiz:
@christophe-g Thanks again. I'll close for now for the reasons cited earlier. I'm sure a contribution like this would be welcomed in the future!

PR #92, opened 2015-03-17
Allow dba to execute queries

joewiz:
The `@execute-query` restriction controls if non-dba users are allowed to execute XQuery code from within eXide. Previously if this restriction was set to `no`, even dba users were prevented from executing queries. This commit allows dba users to execute queries in the circumstance that `@execute-query` is set to `no`.


joewiz:
Thanks, @wolfgangmm!

PR #99, opened 2015-07-20
modified store.xql to support storing of binary html files

JoernT:
This modification requires another existdb PR that adds the store-as-binary function to the xmldb module


joewiz:
I'm just checking in on the status of the dependency you mention.  Is this store-as-binary function available yet?  If not, we can hold off on merging until then.

JoernT:
first the store-as-binary PR still needs to be merged - see https://github.com/eXist-db/exist/pull/727/files

PR #103, opened 2015-09-11
flex layout broken on chrome 45  (linux)

christophe-g:
The flex layout seems to be broken on latest chrome (container spanning way beyond screen size when lots of tabs are open).

This pr should fix it.
Cheers.
C.


joewiz:
+1 if this fixes the same problem I'm experiencing when I have lots of tabs open.

joewiz:
I just tested it by using the revised eXide.css instead of the current one, and it does the trick. Thanks, @christophe-g!

PR #107, opened 2015-10-03
Conform to default collection.xconf

joewiz:


kohsah:
I think its good to have this enabled. Its now the 2nd time I ran into this problem and spent a good part of an hour wondering why RestXQ was not working. Also since the documentation for Restxq isn't complete per se (there is a lot more information in the eXist book now that I checked again).

adamretter:
+1

PR #121, opened 2016-01-15
add adaptive & json output option for query results

joewiz:
## summary
- adds an "Adaptive Output" option alongside the existing "Direct Output" and "XML Output" options for eXide query results
- follows all rules at according to http://www.w3.org/TR/xslt-xquery-serialization-31/#adaptive-output, except we cannot currently support the xs:double rule as it reads (the picture string specified in the spec returns errors in eXist) and xs:NOTATION (I wasn't able to find any information on how this obscure datatype should be serialized). i'll happily add support for these as the picture string situation is improved and as I learn more about xs:NOTATION.
### update (Jan 17, 2016)
- fixed a bug in the display of maps in adaptive output mode containing multiple objects.
- changed display of maps in adaptive output mode to indent each new map & object
- added a "JSON Output" option. this pipes results through `fn:serialize` using JSON serialization output options with indent="yes". relieves users of pasting in the verbose serialization options when working with strictly JSON data
- fixed a typo in the display of anonymous functions
- improved handling of non-anonymous functions
## motivation
- even though this is faux serialization and just another mode of pretty printing, the motivation is similar to that stated in the spec: "to allow any valid XDM instance to be serialized without raising a serialization error"
- rather than serialization errors in eXide's XML Output mode, we more often see silently empty results for maps, arrays, and named/anonymous functions. 
- adaptive output provides a way to represent these structures.
- specifically, adaptive output provides more illuminating results for data types like attributes, xs:QName, string vs. numbers. the results are clearer because attributes are returned in `name="value"` form, strings are surrounded with quotation marks (e.g., `"a"` or `"42"`), and numbers are not (e.g., `42`); and because i've applied basic syntax coloring to the results - borrowing both from ace's default coloring for XQuery as well as the classes used in the XML pretty printing function. the color selections are preliminary and could probably be improved:
## illustrations
- compare the original XML Output:

screen shot 2016-01-15 at 3 05 49 am
- to the new Adaptive Output:

screen shot 2016-01-18 at 11 44 24 pm
- and compare the old boilerplate for serializing JSON and the unindented results:

screen shot 2016-01-17 at 11 45 16 pm
- to the new JSON Output option, with more direct code and indented results:

screen shot 2016-01-17 at 11 57 13 pm
## additional notes
- i've taken pains not to alter eXide's existing behavior for the "XML Output" in any way
- also, when outputting normal XML nodes, eXide uses the same pretty printing function as before, even when in "Adaptive Output" mode
- thus the only change users will encounter when they select "Adaptive Output" mode, is when results include (1) attribute nodes or (2) the non-node XDM data types supported by adaptive output
- i placed "Adaptive Output" as the first option in the drop down menu, because (1) alphabetically it precedes XML and (2) the effect on users is limited since the only change users would notice would be for attribute nodes and non-XDM datatypes. but since this changes the default, I could certainly move it down below "XML Output" if preferred.
- the only other concern is that users may think that adaptive output is supported as a full-blown serialization option (i.e., outside of eXide), but (1) i imagine that this option may be added at some point anyway, and (2) in the meantime such a misunderstanding could be prevented through documentation (release notes, a new "did you know" pop-up, etc.). i think this concern is much smaller, though, than the benefit users would get from this option as they increasingly make use of non-node datatypes. 
## test queries

... from the illustrations above 

``` xquery
xquery version "3.1";

map {"answer": 42},
<question answer="42"/>,
<question answer="42"/>/@answer,
attribute foo {"bar"},
attribute xsi:type {"xs:integer"},
QName("http://www.tei-c.org/ns/1.0", "persName"),
42 = 42,
string-join#1,
function($x, $y) {$x * $y},
[
    map {"answer": 42},
    <question answer="42"/>,
    <question answer="42"/>/@answer,
    attribute foo {"bar"},
    attribute xsi:type {"xs:integer"},
    QName("http://www.tei-c.org/ns/1.0", "persName"),
    42 = 42,
    42,
    -42,
    xs:date('1942-02-02'),
    string-join#1,
    function($x, $y) {$x * $y}
]
```

``` xquery
xquery version "3.1";

declare namespace output="http://www.w3.org/2010/xslt-xquery-serialization";

let $json := json-doc("http://lae.princeton.edu/catalog/0bp35.jsonld")
let $serialization-parameters := 
    <output:serialization-parameters>
        <output:method>json</output:method>
        <output:indent>yes</output:indent>
    </output:serialization-parameters>
return
    serialize($json, $serialization-parameters)
```

``` xquery
xquery version "3.1";

json-doc("http://lae.princeton.edu/catalog/0bp35.jsonld")
```


src/eXide.js
```diff
@@ -571,11 +572,14 @@ eXide.app = (function(util) {
...
 			$.log("retrieveNext: %d", currentOffset);
 		    if (currentOffset > 0 && currentOffset <= endOffset) {
+		        $("#serialization-mode").removeAttr("disabled");
+		        var serializationMode = $("#serialization-mode").val();
 		        var url = 'results/' + currentOffset;
 				currentOffset++;
 				$.ajax({
 					url: url,
 					dataType: 'html',
+					data: { "output": serializationMode },
```
joewiz, line 583: Notice I have to pass the serialization/output parameter to the `/retrieve` endpoint so it knows whether to pretty-print the results as XML Output or Adaptive Output.

dizzzz:
Nice! very extensive....

joewiz:
Thank you @dizzzz!

joewiz:
@wolfgangmm Thanks for accepting the contribution! I hope people find it useful. I have some other ideas for improvements along these lines.

christophe-g:
Thanks a lot for this work ! looks really good ; )

joewiz:
Thanks, that's great to hear, @christophe-g!

PR #129, opened 2016-08-11
Use standard collection uri

joewiz:
The non-standard uri worked, but better to use the standard one.


joewiz:
Will submit PR from my own branch

PR #133, opened 2016-09-08
show <exist:match> element in hits

### LINKED ISSUE #132: Show <exist:match> element in results
As reported by Martin Holmes on exist-open http://markmail.org/message/j3kpq423z2kh23co:

> I've realized that it's eXide that's getting in my way. Using 
> util:expand, I wasn't seeing the exist:match tags in the output at the 
> bottom of the screen, and I couldn't easily select and copy any 
> significant amount of text from there. Switching to the Java admin 
> client, everything is much smoother.

I propose changing the pretty printing module to retain the `<exist:match>` element while still highlighting the contents.

joewiz:
Closes #132


joewiz:
Just to reiterate the content of my commit message just now, I revised this PR with to give users control over the application of `util:expand()`, which was automatically applied to all nodes when the (previously default) "XML Output" option was selected. Ever since adding the "Adaptive Output" mode and making it default, `util:expand()` was not applied automatically to its results. Therefore, results were being handled differently between "XML Output" and "Adaptive Output."

Therefore, to unify the handling of lucene matches across these methods, and to make this behavior explicit, I've updated the PR to include a new checkbox, "Auto Expand Matches", enabled by default, which preserves the current behavior for "XML Output" and applies that default to "Adaptive Output." In this mode, the `<exist:match>` element still appears, as proposed in #132. Unselecting this checkbox, however, disables the automatic application of `util:expand()` to nodes, meaning users would be seeing a more accurate representation of the results of their query. They can still apply `util:expand()` manually, and thus they will be able to learn how to use this function - seeing how it affects results.

To summarize the PR, here is the before & after screenshot, showing the result of a Lucene full text search like this:

``` xquery
xquery version "3.1";

declare namespace tei="http://www.tei-c.org/ns/1.0";

collection('/db/apps/milestones/chapters')//tei:div[ft:query(., 'taiwan')]
```

**Before:**

screen shot 2016-09-07 at 11 35 32 pm

**After:** (Default)

screen shot 2016-09-13 at 12 40 25 pm

**After:** (With `Auto Expand Matches` unchecked)

screen shot 2016-09-13 at 12 40 46 pm

joewiz:
@martindholmes I'd be interested in your take on my revised PR, as described and illustrated in https://github.com/wolfgangmm/eXide/pull/133#issuecomment-246746265.

martindholmes:
This looks perfect for the kind of testing I'm doing. I vote yes!

joewiz:
Great to hear, thanks!

joewiz:
I'll update the conflicts here once #137 is merged.  

I also will clarify the label for the new checkbox that controls whether matches are auto expanded. Instead of "Auto Expand Matches", how about "Highlight Index Matches". The earlier label was technically correct but probably wouldn't make sense to new users. Highlight Index Matches is also correct but should make more intuitive sense to new users. 

The only other possible change I could think of would be to move this into eXide's preferences (to reduce clutter in the query window). But it's not too bad as is, right?

@wolfgangmm Are there any other concerns you have with the changes in this PR?

joewiz:
**Updates**:
- I've rebased to eliminate conflicts from commits that came after this PR was first submitted
- I've changed the wording of the checkbox to read "Highlight Index Matches" instead of (the less clear) "Auto Expand Matches"
- I've added handling for matches on attributes, using the new `util:expand()` as revised in #1135. Here's a screenshot showing the highlighting of the sample code from https://github.com/eXist-db/exist/pull/1135#issuecomment-249383273:

> screen shot 2016-10-28 at 2 01 46 pm

The actual code that produced this result:

``` xquery
xquery version "3.1";

declare namespace tei="http://www.tei-c.org/ns/1.0";

let $clear := 
    for $col in ('/db/test', '/db/system/config/db/test')
    return
        if (xmldb:collection-available($col)) then 
            xmldb:remove($col)
        else
            ()
let $data-collection := xmldb:create-collection('/db', 'test')
let $xconf-collection := xmldb:create-collection('/db/system/config/db', 'test')
let $xconf := 
    <collection xmlns="http://exist-db.org/collection-config/1.0">
        <index xmlns:tei="http://www.tei-c.org/ns/1.0" xmlns:xs="http://www.w3.org/2001/XMLSchema">
            <lucene>
                <text qname="@lemma" type="xs:string"/>
            </lucene>
        </index>
    </collection>
let $store-xconf := xmldb:store($xconf-collection, 'collection.xconf', $xconf)
let $in-memory-doc := 
    document { 
        <TEI xmlns="http://www.tei-c.org/ns/1.0">
            <p>
                <s>
                    <w>KWIC</w><w>highlighting</w>
                    <w>code</w><w>has</w>
                    <w>been</w>
                    <w lemma="success">successfully</w>
                    <w lemma="commit">committed</w>
                    <w>to</w><w>the</w><w>github</w><w>repository</w>!
                </s>
            </p>
        </TEI>
    }
let $store-doc := xmldb:store($data-collection, 'test.xml', $in-memory-doc)
let $on-disk-doc := doc($store-doc)
return
    $on-disk-doc//tei:w[ft:query(@lemma, <query><term>commit</term></query>)]
```

joewiz:
@dizzzz I see you assigned the label "0 - Backlog" - the other labels being "1 - Ready", "2 - Working", and "3 - Done."  What does the "backlog" label mean?  Do you have any suggestions about this PR?  Thanks in advance for any advice you might have.

dizzzz:
hmmm a backlog item for me is: an identified issue that will be planned in during a 'sprint' when the priority is high enough :-) http://scrummethodology.com/the-scrum-backlog/ 

I marked it this way to be sure it has attention :-)  but I realise it dus not make too much sense.

dizzzz:
regarding your PR... the output window looks very nice..... shallI pull in? (you clearly tested it well)

joewiz:
Thanks for the explanation - I wasn't familiar with the terminology. And thanks for assigning a label to keep this PR from being lost. I did test it (a little more just now, in fact), and I think it's ready to merge. Let me sum up the two new features as follows:
1. eXide now highlights the results of index-backed queries on attributes (see https://github.com/eXist-db/exist/pull/1130)
2. eXide has a new checkbox in the query results pane, called "Highlight Index Matches."
   - When enabled (its default state), queries that leverage eXist's full text or range indexes return results in which index matches are highlighted with a red underline. When the new checkbox is disabled, the results index-backed queries are not highlighted. 
   - Long time users of eXide might be a little puzzled, since eXide has long offered automatic highlighting of index-backed results. (As before, it only applies to the Adaptive Output and XML Output modes.) The new feature actually solves a few problems with the handling of query results in eXide. It gives you more control over the appearance of query results, gives you a more accurate representation of query results, and allows users to explore how the `util:expand()` function works from within eXide, which was previously impossible. To expand on each of these: 
   - More control over the appearance of query. Before, eXide applied highlighting to all results. It did this by applying the `util:expand()` function to all results, which causes eXist to apply an `<exist:match>` element to all index matches. eXide turned these `<exist:match>` elements into a red underline style. This was a great default, but eXide didn't allow users to turn off the default highlighting. The new checkbox gives users the ability to disable the default highlighting.
   - More accurate representation of results: Before, eXide the `<exist:match>` element was not shown in query results. This was not a significant problem, since the red underline obviously represented the `<exist:match>` elements. However, it could cause confusion in limited circumstances. If a user invoked the `util:expand()` function in their own queries, they would not see any change in their query results, since eXide suppressed all `<exist:match>` elements. The new feature now displays `<exist:match>` elements, in addition to the red underlining. In this way, users can see the result of `util:expand()` in their queries.
   - Allows users to explore how the `util:expand()` function works: Because eXide stripped `<exist:match>` elements from results, users could not use eXide to learn how to use the `util:expand()` function to control the output of their queries. Learning this is actually a very important technique for getting the most out of eXist's indexes. A newly available mode, available by unselecting the "Highlight Index Matches" checkbox, gives users complete control over the appearance of their query results by disabling the highlighting and letting users experiment with `util:expand()`
   - Thus, this feature preserves the previous behavior of highlighting results, but reveals the previously-suppressed elements, with improved accuracy in the display of results and a more useful learning environment.

Perhaps this text could be used in a blog post announcing the next release of eXide when this is included? I would be happy to write a similar summary of the other features I contributed, "Adaptive Output" and "JSON Output." It could go in the atomic wiki and/or in the eXide documentation (via Help > Documentation) http://localhost:8080/exist/apps/eXide/docs/doc.html - which currently discusses eXide 2.0.8 (though eXide is up to 2.2.0 now). Perhaps the 3.0RC2 / 3.0 final release would be a good time for an updated release of eXide? Just some ideas!

joewiz:
I discovered that the test suite runner (invoked via `XQuery > Run as test`) referenced an earlier version of the `pretty-print()` function signature, so I just fixed that.

modules/session.xql
```diff
@@ -42,11 +42,11 @@ declare function sandbox:retrieve($num as xs:integer) as element() {
     let $cached := session:get-attribute("cached")
     let $cached-item := $cached[$num]
     let $result := 
-        if ($output = ("xml", "adaptive")) then
+        if ($output = "xml") then
```
dizzzz: '=' or 'eq' ?

joewiz: before, with multiple values, it had to be `=`, but with a single value, either works. is there a reason to prefer `eq`?

dizzzz: I have no preferences, but I was wondering... is a value or a sequence comparison intended ?

joewiz: as revised, this is clearly a value comparison, not a general comparison. i will make the change, but there are tens of other cases in pre-existing code just in eXide's modules directory, not to mention other apps. if there are performance benefits, it would certainly be worth applying this principle elsewhere!

joewiz:
Thanks for accepting the PR @wolfgangmm, and for all the suggestions @dizzzz!

Please let me know if you would like me to write anything more than I did above for documentation/release notes.

PR #136, opened 2016-10-27
handle null & empty sequences in adaptive output

joewiz:
before, results including JSON null values (i.e., an empty sequence in the value of a map entry) led to a silent error. now, the handling is as follows:

``` xquery
parse-json(' {"hello": null} ')
```

returns

``` xquery
map { "message": () }
```


joewiz:
Thanks, Wolfgang! I realize when looking at the diff that I didn't commit my final version.  (It returns "null" instead of "()".) I'll submit another fix this morning.

joewiz:
Thanks, Wolfgang! I realize when looking at the diff that I didn't commit my final version.  (It returns "null" instead of "()".) I'll submit another fix this morning.

PR #138, opened 2016-12-05
Fix function doc lookup & mapped module resolution

### LINKED ISSUE #128: Function lookup/validation failures with math functions
Looking up `math:pow` (via Navigate > Function doc or F1) returns 0 results, but looking it up as `pow` yields the documentation for the function, but without its `math` namespace prefix. When selecting this function to autocomplete the function signature, eXide says (correctly) that there is no `pow` function in the default function namespace.

### LINKED ISSUE #139: Problems with function documentation lookup and autosuggested module import
## Problems with function documentation lookup (`docs.xql`)

1. As first noted in #128, looking up `math:pow` (via Navigate > Function doc or F1) returns 0 results, but looking it up as `pow` yields the documentation for the function, but without its `math` namespace prefix. When selecting this function to autocomplete the function signature, eXide says (correctly) that there is no `pow` function in the default function namespace.

    > screen shot 2016-12-06 at 8 04 40 pm

    > screen shot 2016-12-07 at 1 46 35 am

    This problem didn't affect the `map` or `array` functions since the `function/@name` values for these functions contained the module name.

1. I discovered a similar set of issues affecting functions in the default function namespace when looking up these functions using the `fn:` prefix. While not conventionally used in eXist XQuery circles, it is legal to include, e.g., `fn:string-join()`. In eXide, autocomplete failed (i.e., returned `null`) on `fn:`-prefixed functions.

    > screen shot 2016-12-07 at 1 36 09 am

    An additional benefit of supporting `fn:` would be as a filter for limiting search results to those in this namespace; otherwise, these most common of functions are necessarily interspersed with results of all other modules.

1. Only "registered modules" are included in function documentation lookup and function auto-complete. The code for looking up "mapped modules" was commented out in `docs.xql`, and when uncommented, yielded errors in my testing. As a result, looking up functions in the kwic, testing, json, and jsonp modules wasn't possible in eXide. This made it seem as if the functions, despite being mapped in `conf.xml`, were unavailable. (They did appear in the listing of Navigate > Go to/import module, so they could be imported via this method; however, "go to" doesn't work for these modules.)

    > screen shot 2016-12-07 at 1 41 41 am

## Problems with autosuggested module import (`find.xql`):

1. Only "registered modules" are included. "Mapped modules" are excluded. These include kwic, testing, json, and jsonp modules. I frequently use `kwic:summarize()` and then get frustrated that eXide can only auto-suggest this much: `import module namespace kwic="";`

    > screen shot 2016-12-07 at 1 43 08 am
    ... Oh no! Need to import kwic!

    > screen shot 2016-12-07 at 1 43 20 am
    ... Great! eXide is going to construct the import statement for me!

    > screen shot 2016-12-07 at 1 19 11 am
    ... No!!!

joewiz:
Fixes #128

**Update (2016-12-07)**: I closed #128 and opened #139 - an expanded description of the problem. This fixes #139.

### DIFF
--- a/modules/docs.xql
+++ b/modules/docs.xql
@@ -18,14 +18,115 @@
  :)
 xquery version "3.0";
 
-declare namespace xqdoc="http://www.xqdoc.org/1.0";
 declare namespace json="http://json.org/";
+declare namespace xqdoc="http://www.xqdoc.org/1.0";
 
 declare option exist:serialize "method=json media-type=application/json";
 
-declare function local:create-template($signature as xs:string) {
+(: Search for functions matching the supplied query string.
+ : Logic for different kinds of query strings:
+ :   1. Module namespace prefix only (e.g., "kwic:", "fn:"): show all functions 
+ :        in the module
+ :   2. Module namespace prefix + exact function name (e.g., "math:pow"): show 
+ :        just this function
+ :   3. Module namespace prefix + partial function name (e.g., "ngram:con"): 
+ :        show matching functions from the module
+ :   4. No module namespace prefix + partial or complete function name (e.g., 
+ :        "con"): show matching functions from all modules
+ : Note: We give special handling to default XPath functions:
+ :   1. Since the "fn" namespace prefix is the default function namespace, 
+ :        its functions are included in searches when no namespace prefix is
+ :        supplied. Functions from this namespace appear at the top of the list 
+ :        of results. The results also omit the "fn" namespace prefix if it 
+ :        was omitted in the query string. 
+ :   2. If the "fn" namespace prefix is supplied in the query string, we limit 
+ :        searches to the default XPath functions, and the results show the 
+ :        prefix.
+ : :)
+declare function local:get-matching-functions($q as xs:string) {
+    let $supplied-module-namespace-prefix := if (contains($q, ':')) then substring-before($q, ':') else ()
+    let $function-name-fragment := if (contains($q, ':')) then substring-after($q, ':') else $q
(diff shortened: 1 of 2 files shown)


dizzzz (approved):

joewiz:
Thanks @dizzzz. I'd be happy to answer any questions about this PR.

joewiz:
I just rebased this branch to fix an expanded set of issues as described in #139. To summarize the fixes in the revised branch:

- consolidate lookup of registered and mapped modules in both function doc lookup and module import resolution
- fix math module lookup problems (originally #128)
- add support for fn:-prefixed function lookup, and respect the prefix if supplied
- fix resolution of mapped modules for autosuggest module import
- added comments to explain handling of various cases
- general refactoring and reformatting

joewiz:
@dizzzz After your original review I realized I could address a broader set of problems related to the original one. Rather than close and open a new branch & PR I made my changes to the original one. Sorry if this complicates the "approved" labels, etc.!

joewiz:
@wolfgangmm If you think atom-existdb would benefit from any of these changes, I'd be happy to port these to https://github.com/wolfgangmm/atom-editor-support/blob/master/atom-autocomplete.xql. While I think you've already improved that over what's in eXide, I think it still suffers from the problem with `math` module, doesn't support `fn` prefixed functions, and isn't able to look up `kwic` or other mapped modules.

joewiz:
Thanks for the review @dizzzz and for the merge @wolfgangmm.

PR #140, opened 2016-12-09
Better Less

line-o:
- update less.js to 2.5.1
- improve working with partials
  - partial files start with an underscore by convention
  - they will not be transformed into a css file
- generated CSS files are prepended with a comment/ warning not to edit this file directly


dizzzz (approved):

dizzzz:
@wolfgangmm  ?

PR #142, opened 2017-01-23
Restrict App Access

line-o:
When setting guest access to "no" in configuration.xml, a guest is denied access to the index page of the eXide application.
screen shot 2017-01-23 at 12 09 16

The check is added to the controller.

### DIFF
--- a/controller.xql
+++ b/controller.xql
@@ -24,7 +24,7 @@ declare variable $login :=
         else
             local:fallback-login#3
 ;
- 
+
 (:~
     Fallback login function used when the persistent login module is not available.
     Stores user/password in the HTTP session.
@@ -39,7 +39,7 @@ declare function local:fallback-login($domain as xs:string, $maxAge as xs:dayTim
             error(xs:QName("login"), "Persistent login module not enabled in this version of eXist-db")
         else if ($logout) then
             session:invalidate()
-        else 
+        else
             if ($user) then
                 let $isLoggedIn := xmldb:login("/db", $user, $password, true())
                 return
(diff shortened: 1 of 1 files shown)


controller.xql
```diff
@@ -137,15 +137,23 @@ else if (starts-with($exist:path, "/check/")) then
...
 else if ($exist:resource = "index.html") then
-    <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
-        <view>
-            <forward url="modules/view.xql">
-                <set-header name="Cache-Control" value="max-age=3600"/>
-            </forward>
-        </view>
-    </dispatch>
+    return
+        if (local:user-allowed())
```
rvdb, line 143: This makes the eXide login screen unavailable to unauthenticated users, thus making it impossible to log in via eXide. Is this the purpose of this check? If so, I presume users are supposed to log in via e.g. the dashboard, before they can access eXide (if they're allowed).

If so, I think the `local:user-allowed()` function should check against session attributes instead of request attributes. Since no authentication is possible in the request to eXide, the function will never evaluate to `true()` in its current form:

    declare function local:user-allowed() {
      (
        request:get-attribute("org.exist.login.user") and
        request:get-attribute("org.exist.login.user") != "guest"
      ) or config:get-configuration()/restrictions/@guest = "yes"
    };

Instead, if it would check against session attributes, that would make it possible to honour the authentication status set in other apps such as the dashboard:

    declare function local:user-allowed() {
      (
        session:get-attribute("org.exist.login.user") and
        session:get-attribute("org.exist.login.user") != "guest"
      ) or config:get-configuration()/restrictions/@guest = "yes"
    };

PR #143, opened 2017-01-24
consolidate branches

line-o:
* adds all commits in master that were missing from develop
* bumps version to 2.3.0

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -1,1 +1,1 @@
-project.version=2.2.0
+project.version=2.3.1
--- a/build.properties
+++ b/build.properties
@@ -1,1 +1,1 @@
-project.version=2.3.1
+project.version=2.3.0
--- a/build.xml
+++ b/build.xml
@@ -1,7 +1,6 @@
 <project basedir="." default="all" name="eXide">
     <property file="local.build.properties"/>
     <property file="build.properties"/>
-    <property name="project.version" value="1.0.16"/>
     <property name="src" value="./src"/>
     <property name="tools" value="./tools"/>
     <property name="build" value="./build"/>
--- a/docs/doc.html
+++ b/docs/doc.html
@@ -12,46 +12,49 @@
         </div>
     </div>
     <section id="news">
-        <h1>New in 2.0.8</h1>
+        <h1>New in 2.3.0</h1>
         <div class="feature">
             <div class="row-fluid">
                 <div class="span3 thumb">
-                    <h2>Live Reload</h2>
-                    <p>Instant previews</p>
+                    <h2>Adaptive &amp; JSON Output</h2>
+                    <p>New output modes for viewing your query results</p>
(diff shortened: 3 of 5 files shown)


joewiz (commented):
@line-o These changes look good. I'm curious: is there a reason you're bumping to 2.3.1 and not 2.3.0?  Also, I have some notes to add to repo.xml describing the additions in this release. I could submit these as part of this PR or as a separate one - what would you prefer?

joewiz:
The repo.xml addition would be as follows (replacing 2.3.1 with 2.3.0 if you agree that 2.3.0 is better, since there was no 2.3.0 release to my knowledge):

```xml
        <change version="2.3.1">
            <ul xmlns="http://www.w3.org/1999/xhtml">
                <li>New adaptive and JSON output options for query results</li>
                <li>New checkbox to toggle highlighting of index hits</li>
                <li>Fix function doc lookup</li>
                <li>Various small bug fixes</li>
            </ul>
        </change>
```

Again, I'm happy to submit this in this or in a separate PR.

line-o:
@joewiz 2.3.0 is the current one in the package repository. 
The additions to `repo.xml` should be part of this one. Good that you noted it, as I did not.

line-o:
👍

joewiz:
@line-o Ah, I see. Something strange is going on though, since I've never seen 2.3.0 via the package manager pane in Dashboard. Indeed, even running the latest nightly, the package repository still shows 2.2.0 to me: 

> screen shot 2017-01-25 at 7 45 22 am

(This is showing what's available from the repo, not what's installed on my machine.)  

The package manager's request that returns this entry is: http://localhost:8080/exist/apps/dashboard/plugins/packageManager/packages/?format=manager&type=remote.  I wonder if this request isn't properly sending the version parameter that the package manager relies on to determine compatibility. I've fixed some aspects of the public-repo and dashboard, but I don't think I've fixed this request. I do know that the package repo (at least the one on exist-db.org) has some bugs that I've fixed in source but aren't running on exist-db.org. I'll look into this package manager request bug later this morning.

**Update**: There was no bug in the package manager; as noted below, the problem was in the contents of the public repo.

joewiz:
Wait, take a look at http://demo.exist-db.org/exist/apps/public-repo/packages/eXide.html?eXist-db-min-version=3.0.0 - 2.3.0 is not listed.  Also, the 2.3.0 version number was not in source - as the diff of this PR shows. So if 2.3.0 was in the public-repo, I don't think the listing of apps was ever refreshed (a manual step separate from uploading the .xar), which would've updated the listing on this URL.  So no one could've installed it.  In light of this, I think 2.3.0 was never "released" - so by that logic, should the version number should be switched to 2.3.0?

line-o:
You are right, v2.3.0 it is then.

PR #158, opened 2017-03-15
set expand-xincludes=no when downloading app xar

### LINKED ISSUE #90: Disable xinclude processing in eXide's "download app" command
Moved from https://github.com/eXist-db/exist/issues/403, where @awagner-mainz suggested (a) disabling expansion of XIncludes in eXide's "download app" or (b) at least have an option to turn off XInclude expansion, and added:

> After running into this problem for the fiftieth time (or so it feels), I have again lost my xi:include statements somewhere along the download/edit/distribute/deploy path and ended up with flat (and redundant) xml files. On the eXist-db mailing list, other persons have confirmed to have run into this.

joewiz:
- exposes a new request parameter, expand-includes, which is now set to “false” by default but could be set to “true” by the application
- closes #90

### DIFF
--- a/modules/deployment.xql
+++ b/modules/deployment.xql
@@ -21,6 +21,7 @@ xquery version "3.0";
 
 import module namespace apputil="http://exist-db.org/apps/eXide/apputil" at "util.xql";
 import module namespace tmpl="http://exist-db.org/xquery/template" at "tmpl.xql";
+import module namespace dbutil="http://exist-db.org/xquery/dbutil";
 
 (:~ 
     Edit the expath and repo app descriptors.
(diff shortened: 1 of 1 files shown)


modules/deployment.xql
```diff
@@ -571,9 +572,24 @@ declare function deploy:package($collection as xs:string, $expathConf as element
...
+                    <entry name="{$relative-path}" type="uri">{$resource}</entry>
+                else
+                    <entry name="{$relative-path}" type="xml">
+                        {
+                            util:declare-option("exist:serialize", "expand-xincludes=" || (if ($expand-xincludes) then "yes" else "no")), 
+                            doc($resource)
+                        }
+                    </entry>
+        })
+    let $xar := compression:zip($entries, true())
```
dizzzz, line 592: so in essence in stead of bundling a whole collection, individual resources are bundled. Why not :-)
How are collections bundled in the ZIP? implicitly because a resource is referenced? 

I see your remark about empty collections, so this PR implies a behavioural change.....

joewiz: Thanks @dizzzz! 

Correct, we just do individual resources, and their by storing the resources' relative path, the .zip file, when uncompressed, will create the necessary directory structure. 

In my original approach, I used `dbutil:scan` and handled collections too, not just resources, but I noticed that given an empty collection like `/db/some-collection`, including `<entry name="/db/some-collection" type="collection">` didn't deposit this empty directory into the .zip file. I didn't investigate further whether this was a limitation of `compression:zip` or the zip file format or what, but I thought I'd note the issue here.

joewiz:
@dizzzz As no one else has weighed in, could you please approve/merge this when you have the chance? I would be happy to prepare a new release in advance of the 3.2.0 release, so this fix could go out to everyone. By the way, in https://github.com/wolfgangmm/eXide/issues/90#issuecomment-286784439 I described how I tested the PR - it works for me as described here.

dizzzz:
The PR needs to be reworked it seems

joewiz:
@dizzzz Thx!

PR #160, opened 2017-04-13
Fix/go to line column offset

### LINKED ISSUE #159: Navigate > Go to line > Column goes 1 character too far
The Navigate > Go to line dialog allows the user to specify a line and column at which to place the cursor. This dialog takes users to the correct line, but the wrong column number. It takes you one column too far. For example, given a line:

```text
"hello"
```

Going to column 1 should place the cursor just before the 1st character (the quotation mark `"`). Going to column 2 should place the cursor just before the 2nd character (the letter `h`). 

But in eXide, if you request column 1, it'll place the cursor just before the 2nd character.

As a result, when an error message says that an error takes place on line 7 column 9 with an unexpected token `element`, and you use the Go to line dialog to go to this location, the cursor is placed after the first `e` instead of before it.

joewiz:
Closes #159.


dizzzz (approved):

joewiz:
Thanks for the review and merge @dizzzz!

Issue #34, opened 2013-09-04
Select tab with file opened with 'open' param


wshager:
When a file is opened using ?open=/db/file AND the previous state of eXide is restored AND the file was already open in that state, the tab containing the file is not necessarily selected. This looks like an issue, however minor.

wolfgangmm:
Fixed in e64ad30717b665fa8108e6d3bf3cd299bfd01392

PR #162, opened 2017-04-19
Serialize strings when showing query results

### LINKED ISSUE #161: Serialize strings rather than returning literals during pretty-printing
As reported by @Torcsi in https://github.com/eXist-db/exist/issues/1420 and demonstrated by @dizzzz, given an element like `<a b="&amp;#34;c"/>`, eXide returns this as

> 3.1.x / eXide returns `<a b="&#34;c"/>`

whereas elsewhere this is returned with the ampersand properly escaped:

> 3.1.x / query via REST returns `<a b="&amp;#34;c"/>`
> 3.1.1 / doc retrieval via REST returns `<a b="&amp;#34;c"/>`

It appears that in eXide we return the literal string instead of serializing these strings. As a result, the browser is displaying `&amp;` as `&`. Applying `fn:serialize()` to the strings should cause eXist to return `&amp;amp;#34;c` which the browser will then display as the expected result, `&amp;34;c`.

joewiz:
Closes #161 

Now, the `&amp;` in`<a b="&amp;#34;c"/>` is returned intact rather than being returned as `<a b="&#34;c"/>`), whether the default Adaptive Output or XML Output is selected.

Also:

- `"&amp;"` returns the same instead of `"&"`
- `<x>&amp;</x>` returns the same instead of `<x>&</x>` 
- `<x xmlns="http://foo.org?a=1&amp;b=2"/>` returns the same, instead of `<x xmlns="http://foo.org?a=1&b=2"/>`

Tested this behavior in BaseX (which has implemented adaptive serialization) with queries like this:

```xquery
xquery version "3.1";

declare namespace output="http://www.w3.org/2010/xslt-xquery-serialization";

declare option output:method "adaptive";

"&amp;"
```

### DIFF
--- a/modules/pretty-print.xql
+++ b/modules/pretty-print.xql
@@ -90,7 +90,7 @@ declare function pretty:pretty-print-xml($node as item(), $namespaces as xs:stri
                                 
                         }
                         </span>, '="',
-                        <span class="ace_string">{$nsDecl/@uri/string()}</span>,
+                        <span class="ace_string">{serialize($nsDecl/@uri/string())}</span>,
                         '"'
                     )
                 }
@@ -109,7 +109,7 @@ declare function pretty:pretty-print-xml($node as item(), $namespaces as xs:stri
                                 ),
                                 " "
                             )
-                        }">{$attr/string()}</span>, '"'
+                        }">{serialize($attr/string())}</span>, '"'
 					)
 				}
 				{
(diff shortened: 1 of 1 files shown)


dizzzz (approved):

joewiz:
@dizzzz Thanks for the approval. 

Would anyone else like to weigh in? If not, we'll merge this in soon. I think we could prepare a small release with this and https://github.com/wolfgangmm/eXide/pull/160.

joewiz:
@dizzzz As no one else has weighed in, could you please merge this when you have the chance? I would be happy to prepare a new release in advance of the 3.2.0 release, so this fix could go out to everyone.

joewiz:
@dizzzz Thx!

PR #166, opened 2017-05-31
Ensure download app descends into collections when building xar

### LINKED ISSUE #165: Use "Download app" item in the Application menu, sub-collections in a .xar are lost.
Test for Java 1.8.0_131 + eXist-db 3.1/3.2/3.3 + eXide 2.3.3

joewiz:
Closes #165.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -1,1 +1,1 @@
-project.version=2.3.3
+project.version=2.3.4
--- a/modules/deployment.xql
+++ b/modules/deployment.xql
@@ -575,16 +575,17 @@ declare function deploy:package($collection as xs:string, $expathConf as element
 declare function deploy:download($app-collection as xs:string, $expathConf as element(), $expand-xincludes as xs:boolean) {
     let $name := concat($expathConf/@abbrev, "-", $expathConf/@version, ".xar")
     let $entries :=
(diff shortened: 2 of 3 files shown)


dizzzz:
I have the impression that storing collections by `compression:zip` is only about adding one statement  as the [putEntry](https://github.com/eXist-db/exist/blob/develop/extensions/modules/src/org/exist/xquery/modules/compression/ZipFunction.java#L107) function of of the ZIP class supports. There [actually code](https://github.com/eXist-db/exist/blob/develop/extensions/modules/src/org/exist/xquery/modules/compression/AbstractCompressFunction.java#L309) that works on a "collection" entry, and does perform some magic ; all magic except calling a store function.  I'd recommend to add a `<entry>` for a collection right now in your PR (it is ignored), and have this store collection implemented in the extension function. @adamretter @ljo  do you recall any reason not to store directory/collection items?

dizzzz (approved):
looks good, but I'd suggest to add a `<entry name="{$relative-path}" type="collection">` on the appropriate place for future usage

joewiz:
Thanks for your review and suggestion, @dizzzz, which I incorporated. I've done further testing and can attest that this PR thoroughly fixes the problem. If the PR still looks good to you, would you please squash and merge - if that option is available to you, otherwise just merge? Then I can prepare a release.

PR #170, opened 2017-07-13
Restore setUID, setGID, and sticky bit settings

joewiz:
This feature was originally added in https://github.com/wolfgangmm/eXide/pull/44/commits/909a568a2bd8e35841042c84abe868d321d2a689 but was disabled in https://github.com/wolfgangmm/eXide/commit/154743b5c8cab5620834f0d57f686bad68057543 with the following comment:

> Comment out setuid, setgid bits in form until they are officially supported in eXist.

This was in November 2013. eXist 2.2 (released in November 2014) included the setUID, setGID, and sticky bit feature (see release notes at http://exist-db.org/exist/apps/wiki/blogs/eXist/eXist22). So I think it's okay to re-enable these checkboxes for setting setUID, setGID, and sticky bits.

I also tested the feature and confirmed that it works: setting these permissions in eXide is immediately reflected in the java admin client. And eXide correctly displays the permissions in question (though it always has).

### DIFF
--- a/modules/collections.xql
+++ b/modules/collections.xql
@@ -373,7 +373,7 @@ declare %private function local:get-permissions($perms as xs:string) {
                 <label for="ox">execute</label>
             </td>
         </tr>
-        <!--tr>
+        <tr>
             <td>
                 { local:checkbox("us", substring($perms, 3, 1) = ("s", "S")) }
                 <label for="us">setuid</label>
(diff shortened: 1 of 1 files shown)


dizzzz:
could you add a screen shot of the new situation?

dizzzz (approved):

joewiz:
@dizzzz Sure!  Here it is - the change here affects only the Resource/Collection Properties window within the File > Manage ("DB Manager") window - notice the bottom row of checkboxes:

> screen shot 2017-07-13 at 4 24 33 pm

PR #183, opened 2018-04-05
Removed path names encoding, closes #151

### LINKED ISSUE #151: eXide does query encoding on path names, which is wrong
Steps to reproduce:
- In dashboard load collections
- create collections "ABC" and "A=B" in the root
- store some resources there, eg. "test.html"
- try to edit them in eXide
- /ABC/test.html will open
- /A=B/test.html will open, but one will not see it in the directory panel and one cannot open it from inside of eXide (pun intended)

Looking under the hood of eXide, links below, I see:
- http://localhost:8080/exist/apps/eXide/modules/collections.xql?root=/db/ABC&view=r
- http://localhost:8080/exist/apps/eXide/modules/collections.xql?root=/db/A=B&view=r

Looking under the hood of Dashboard and REST I see:
- http://localhost:8080/exist/apps/dashboard/plugins/browsing/contents/?collection=/db/A=B
- http://localhost:8080/exist/rest/db/A=B

From what I have learned, the call to xmldb:encodeURI at the very end of collections.xql is at fault. I do not have the means to debug further.

obujor:
I've done some testing with special characters in collection names and files and it works.

### DIFF
--- a/modules/collections.xql
+++ b/modules/collections.xql
@@ -47,7 +47,7 @@ declare function local:collections($root as xs:string, $child as xs:string,
             if (sm:has-access(xs:anyURI($root), "r")) then (
                 <title>{xmldb:decode-uri(xs:anyURI($child))}</title>,
                 <isFolder json:literal="true">true</isFolder>,
-                <key>{xmldb:decode-uri(xs:anyURI($root))}</key>,
+                <key>{xs:anyURI($root)}</key>,
                 <writable json:literal="true">{if ($canWrite) then 'true' else 'false'}</writable>,
                 <addClass>{if ($canWrite) then 'writable' else 'readable'}</addClass>,
             	if (exists($children) ) then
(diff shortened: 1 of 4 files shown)


joewiz:
@obujor Wonderful, thanks! Could you please list here the resource and collection names you used in your testing?

duncdrum:
While potentially awesome, I m a bit worried that the default controller will expect url-encoded paths for collections as part of …  urls. Or that `xmldb:create-collection` starts behaving differently when called from inside exide, as opposed to externally

Should there be a test in there to  check 
```xquery
if $path eq xmldb:encode-uri($path)
then …
else …
```
?

joewiz:
@duncdrum Thanks, I'll be sure to consider this in my review. (I don't anticipate those being a problem but will confirm.)

obujor:
@joewiz I don't have all the list (because I deleted some of them) but it was collections and files containing "@" such as these:
- demo@aa.com
- aaa@avavba
- aaaaaa@@@
- bbb

joewiz (approved):
After testing this PR I agree this is a step forward. Aspects of eXide's handling of resource names remain lacking. For example, deletion of resources like `AéB` fails unless we change https://github.com/eXist-db/eXide/blob/develop/src/resources.js#L473 to utilize the resource's `key` instead of its `name`. However, that is outside the scope of this work. At least it is beyond my javascript abilities to carry out thoroughly for all collection/resource actions handled in resources.js. I've noted this in #185.

Issue #185, opened 2018-04-13
Resource operations using "name" instead of "key" are prone to fail


joewiz:
In https://github.com/eXist-db/eXide/blob/develop/src/resources.js, which is responsible for taking user input for operations on resources and crafting HTTP requests to https://github.com/eXist-db/eXide/blob/develop/modules/collections.xql for completing those requests, the js code frequently conflates a resource's `name` (used for user-facing display) with its `key` (used for under-the-hood identification of resources). As a result, operations can fail depending on if the resource name contains URL-encoded characters.

For example, when using the File > DB Manager dialog to delete a collection created via `xmldb:create-collection("/db", "AéB")`, eXide will return a "Deletion of the following items failed: AéB" error, because it crafted the following URL: 

http://localhost:8080/exist/apps/eXide/modules/collections.xql?remove%5B%5D=A%C3%A9B&root=%2Fdb

Instead, it should have crafted this:

http://localhost:8080/exist/apps/eXide/modules/collections.xql?remove%5B%5D=%2Fdb%2FA%25C3%25A9B&root=%2Fdb

The relevant code for deleting resources is https://github.com/eXist-db/eXide/blob/develop/src/resources.js#L473 (changing `.name` to `.key`), but an ideal comprehensive fix would review all resource operations—not just resource deletion—and ensure that the code properly distinguishes between `key` and `name`.

hungerburg:
From eXide 2.4.5 "manage" dialogue I can create a resource/collection with a + sign in its name, then later delete it, but not browse it. Also, the name is shown with the plus as a space.

https://github.com/eXist-db/eXide/blob/bbb84da166a98a91d3fdc3be6e32096afe457ca0/modules/collections.xql#L146

"name" is treated as a _query string_ while it is a _path_; Decoding query strings is not idempotent : %2b → + → space; There is no xmldb function to decode paths; Possibly do a replace ahead of decoding(s/+/%2b/g)?

hungerburg:
@joewiz You give an example of how requests to the backend should be crafted. In my opinion, you have it wrong: One should **not** double encode, never ever; that only will give headaches. Also, the frontend should not be bothered with what encoding the filesystem uses.

joewiz:
[...]

PR #187, opened 2018-04-13
Switch resource handling from name to key for deleteResource

joewiz:
As described at #185. So far, only fixes the deleteResource function. Ideally, this PR would be extended to change all resource functions to use `key` instead of `name`.


joewiz:
I propose merging this now, and extending this work in a future PR.

joewiz:
@dizzzz Thank you!

PR #188, opened 2018-04-14
Do not change owner of /db/apps when creating new app

wolfgangmm:
Fixes https://github.com/eXist-db/exist/issues/1820

### DIFF
--- a/modules/deployment.xql
+++ b/modules/deployment.xql
@@ -140,9 +140,10 @@ declare function deploy:mkcol-recursive($collection, $components, $userData as x
             else
                 "rwxr-x---"
         let $newColl := xs:anyURI(concat($collection, "/", $components[1]))
+        let $exists := xmldb:collection-available($newColl)
         return (
             xmldb:create-collection($collection, $components[1]),
-            if (exists($userData)) then (
+            if (exists($userData) and not($exists)) then (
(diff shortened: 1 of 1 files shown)


joewiz:
The fix makes sense: before, the deploy:mkcol-recursive function was applying the package's required permissions to all ancestor-or-self collections, including /db. Now, permissions aren't set on any collections that already existed before the app was installed. 

@wolfgangmm But this code is only called when eXide creates a new app, right? This wouldn't affect installation of apps via eXide's install-xar-on-upload function, right? For that, eXide and the "classic" dashboard's package manager rely on `repo:install-and-deploy-from-db` (calling it directly [as eXide does](https://github.com/wolfgangmm/eXide/blob/develop/modules/deployment.xql#L604-L612) or letting apputil call it [as dashboard does](https://github.com/eXist-db/shared-resources/blob/master/content/apputil.xql#L139-L169). For that reason, @dizzzz's report is still puzzling to me, and if confirmed suggests that something in https://github.com/eXist-db/exist/tree/develop/extensions/modules/src/org/exist/xquery/modules/expathrepo is amiss?

dizzzz:
IT can be true that I created the XAR file on the same database on which I created the file .....

PR #189, opened 2018-04-15
fix link to emmet.js

joewiz:
emmet.js is deposited directly into shared-resources/scripts, not into the ace subdirectory.


joewiz:
Fixes this error:

> screen shot 2018-04-14 at 8 56 39 pm

joewiz:
@wolfgangmm Thank you!

duncdrum:
shouldn't we also update the dependency declaration for shared-resources in expath-pkg to from 0.4.0 to 0.4.2:
`<dependency package="http://exist-db.org/apps/shared" semver-min="0.4.2"/>`

0.4.2 is current, while I have no idea when and if the resource path changed for emmet, keeping up with the current release seems like a good idea

joewiz:
@duncdrum Good thinking, but in fact, 0.4.0 was the version where this change to the path where we store emmet.js was made: https://github.com/eXist-db/shared-resources/commit/66c56b0df40dbac4c65930f167410720ff3d98b9#diff-2cccd7bf48b7a9cc113ff564acd802a8. So besides the fact that the link to emmet.js has been broken in eXide's develop branch for just under 4 years (!), this tells us there's no immediate need to update eXide's dependency on shared-resources to 0.4.2 (or now 0.5.0), but I'll make a point to review the dependencies before the next release.

PR #195, opened 2018-06-19
Force MSIE to use modern renderer

dizzzz:
In some conditions (corporate policies) the MSIE browser is forced by configuration  to use an old rendering engine.
By setting "X-UA-Compatible" we can allow the browser to use the latest available engine.

https://stackoverflow.com/questions/14611264/x-ua-compatible-content-ie-9-ie-8-ie-7-ie-edge#14611323

### DIFF
--- a/templates/existdb/templates/page.html
+++ b/templates/existdb/templates/page.html
@@ -4,6 +4,7 @@
         <title data-template="config:app-title">App Title</title>
         <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
         <meta data-template="config:app-meta"/>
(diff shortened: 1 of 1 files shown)


joewiz (approved):

joewiz:
@dizzzz Great idea, thanks!

dizzzz:
maybe.... we should define the default browser caching strategy as well.... sometimes I see a page not refreshed whilst it should.... At least I have this impression

joewiz:
In eXide? I don’t think I’ve encountered stale pages, but if you do again, please do share more info about it.

dizzzz:
ah no, I wanted to suggest to add this to the templates that are used to generate a new app. (the change of this PR)

PR #196, opened 2018-06-23
Fix Error 400 when displaying binary result

### LINKED ISSUE #171: Provide better feedback when query returns binary data
Executing the following queries results in silent errors - eXide says 1 result found but shows no result:

- `util:binary-doc("/db/apps/frus-dates/001708346.pdf")`
- `hc:send-request(<hc:request href="http://localhost:8080/exist/apps/frus-dates/001708346.pdf" method="get"/>)[2]`

I had thought that eXide simply wasn't displaying useful feedback about the results being binary data, but in fact the request to `/eXide/results/1` returns a `400 Bad Request` error, with the following response body:

```xml
<exception>
    <path>/db/apps/eXide/modules/session.xql</path>
    <message>exerr:ERROR Unable to encode string value: The underlying InputStream has been closed [at line 37, column 32, source: /db/apps/eXide/modules/session.xql]
In function:
	local:retrieve(xs:integer) [104:24:/db/apps/eXide/modules/session.xql]</message></exception>
```

In general, we should catch such errors and display them to the user. But in this specific case, it would be ideal if we could say `binary results cannot be displayed in eXide`.

joewiz:
Before this PR, submitting a query such as:

```xquery
util:binary-doc("/db/apps/eXide/controller.xql")
```

... would store the binary resource in the session, and eXide would try to fetch this result (via `http://localhost:8080/exist/apps/eXide/results/1?output=adaptive&auto-expand-matches=false&indent=true`), which would return an Error 400 Bad Request:

```xml
<exception>
    <path>/db/apps/eXide/modules/session.xql</path>
    <message>exerr:ERROR Unable to encode string value: The underlying InputStream has been closed
        [at line 37, column 32, source: /db/apps/eXide/modules/session.xql] In function:
        local:retrieve(xs:integer) [104:24:/db/apps/eXide/modules/session.xql]</message>
</exception>
```

Now the following note is substituted for the value:

```
"[This placeholder represents a binary value, which eXide is unable to display.]"
```

Closes https://github.com/eXist-db/eXide/issues/171.

### DIFF
--- a/modules/session.xql
+++ b/modules/session.xql
@@ -48,7 +48,9 @@ declare function local:retrieve($num as xs:integer) as element() {
         </output:serialization-parameters>
     let $serialized :=
         serialize(
-            if ($cached-item instance of node() and $auto-expand-matches) then
+            if ($cached-item instance of xs:base64Binary) then
+                "[This placeholder represents a binary value, which eXide is unable to display.]"
+            else if ($cached-item instance of node() and $auto-expand-matches) then
                 util:expand($cached-item, "highlight-matches=both")
             else
                 $cached-item,


dizzzz:
the good thing, is that this PR results nicely in a predicable result.
however... I am not sure yet if it masks a real issue... @adamretter  ?

dizzzz (approved):

dizzzz:
Probably we don't have a mime-type available, as it would be possible e.g. to render "text/plain" or "text/*" types without to much difficulties (wrap into an XML element when needed)

dizzzz:
for sure a pdf cannot be rendered.

joewiz:
Good idea to probe the mime type, Dannes. I will see if I can at least read and then display the mime type in the comment when available. Also, I hadn’t considered whether a deeper issue is being masked; I’ll see if I can shed any light on the too.

joewiz:
@dizzzz Ok, I've looked into the `Unable to encode string value: The underlying InputStream has been closed` error. To reproduce this error, all you have to do is store a binary value in a session attribute, then try to retrieve it in a subsequent query. 

**test1.xq**

```xquery
xquery version "3.1";

let $binary-value := util:string-to-binary("hello")
let $set-attribute := session:set-attribute("test", $binary-value)
let $attribute := session:get-attribute("test")
return
    <result>{util:binary-to-string($attribute)}</result>
```

... returns:

```xml
<result>hello</result>
```

**test2.xq**

```xquery
xquery version "3.1";

let $attribute := session:get-attribute("test")
return
    <result>{util:binary-to-string($attribute)}</result>
```

... returns:

```xml
<exception>
    <path>/db/test2.xq</path>
    <message>exerr:ERROR The underlying InputStream has been closed [at line 5, column 14]</message>
</exception>
```

No error is logged in exist.log. In contrast, storing the string "hello" and not its binary value results in the 2nd query working without error. It's the binary-ness of the value that is the determining factor in whether an error is thrown.

@adamretter Is this to be expected? Or should I file this as an issue?

joewiz:
@dizzzz I've also confirmed we have no way to retrieve the mime type, since `xmldb:get-mime-type` only works on resources stored in the database. While checking this I realized I should move this PR's code change to where results are serialized, rather than where they're stored in the session. I also changed the note from a comment node to a string, because when "Text Output" is selected, the comment was dropped invisibly; with the new way, though, the note is shown regardless of serialization method.

adamretter:
@joewiz When the query finishes, all binary resources will be closed. Which is why it doesn't survive the session. This is a bug for sure. However, I don't know why you would want to put a binary in the session, the session should be kept as small as possible.

joewiz:
@dizzzz Thanks for the review! Could you merge it please?

line-o:
@joewiz I would like to approve and merge but one question keeps bugging me:
I have opened PDFs in eXide just to see the contents of this binary file (evidence it is not empty). Would this still be possible after your changes are in?
I am actually not depended on that behaviour but I would like to understand the implications first.

joewiz:
@line-o Yes, this PR doesn’t affect opening binary files; it only affects the handling of user-submitted queries that return binaries.

PR #199, opened 2018-08-01
Serve XHTML5 with correct MIME type

dariok:
This PR is related to #198

## Steps taken
1. dropped the `doctype` from the template:
    There is an ongoing discussion about the `doctype` handling in eXist – most recently, see also eXist-db/exist#2056. Specs say that the `doctype` is not a mandatory part of an HTML5 file in XHTML syntax: “XML documents may contain a DOCTYPE if desired, but this is not required to conform to this specification.” (https://www.w3.org/TR/html5/the-xhtml-syntax.html#the-xhtml-syntax). Hence, the `doctype` was removed from the XHTML5 template to avoid confusion.
1. determine MIME type to use based also on the template used
    - A document has a property `template` which is currently not used; the template which was used to create a document is now stored in this property.
    - `template` is then used to determine the MIME type to use for files in the “html“ syntax. Currently, the check is just to see whether the template “xhtml5” has been used, resulting in a MIME type of `application/xhtml+xml` and `text/html` otherwise.

## Further use of changes
By storing the template's name and including it when determining the MIMIE type, additional templates may result in different MIME types, e.g. an HTML file with RDF-A or a TEI-XML as `application/tei+xml`.

## Tests
Tested creating from template an XHTML file, HTML5 fragment, JavaScript, XQuery and XML files and storing them.
MIME type upon retireval is as expected. MIMIE type is correctly loaded and stored again when later opening a file in eXide.
No detrimental effects could be found.


templates/documents.xml
```diff
@@ -16,8 +16,10 @@
     </template>
     <template name="xhtml5" mode="html">
         <description>HTML5 Document in XHTML Namespace</description>
+        <!-- removed doctype: not required per spec
+            (https://www.w3.org/TR/html5/the-xhtml-syntax.html#the-xhtml-syntax).
+            2018-07-31 dariok – https://github.com/dariok -->
         <code>&lt;?xml version="1.0" encoding="UTF-8"?&gt;
-&lt;!DOCTYPE html&gt;
```
dizzzz, line 20: Isnt this required for MSIE ? [I have no idea, bells in the distance]

dizzzz:
Thank you for your PR !

templates/documents.xml
dariok: Well, as it is currently just dropped by eXist, there are not going to be more problems than we have now. I agree that, once this has been dealt with, we should discuss including it again.

However, I do not know how IE 11 deals with this. Older versions of IE may have problems but frankly I would not worry about browsers that have been outdated for more than 5 years. I can check IE11's behaviour later.

duncdrum:
@joewiz can we get a review for this PR?

joewiz (approved):
I've tested the PR and confirmed that creating an HTML document (via New Document > Type > HTML) and saving it to the database results in a stored and served mime type of `application/xhtml+xml`. Looks good.

joewiz:
@dariok Thank you very much for the contribution! We will include this in the next release of eXide.

dariok:
@joewiz and all: glad I could help :)

PR #202, opened 2018-09-09
Fix copy after cut

RemiKoutcherawy:
A cut sets  this.clipboardMode = "move"; then calls copy()
But when doing a Copy after a Cut, copy() never sets this.clipboardMode = "copy"; and performs "move"".
You cannot just add this.clipboardMode = "copy"; in copy() as it will break cut().


joewiz:
Superseded by #225.

PR #212, opened 2019-01-10
Treat XSLT files as XML files - #211

joewiz:
Before this PR, when opening an XSLT file in eXide, eXide would treat the file as XQuery. This PR ensures XSLT files are treated as XML files, so they can benefit from element folding. 

Improves the situation described in https://github.com/eXist-db/eXide/issues/211#issuecomment-452990570, but falls short of requests for XSLT-specific features.


duncdrum (approved):
certainly a step in the right direction. Thx Joe

joewiz:
@duncdrum Thanks!

PR #215, opened 2019-02-15
replace hipchat with slack

duncdrum:
see exist-db/exist#2475


joewiz (approved):

joewiz:
It’s odd that this change would result in a CI failure.

duncdrum:
nope `:latest` is broken, so is develop in the main repo, so forward looking tests are failing

duncdrum:
hmm since this is a readme only change, i m a bit at a loss as to why the tests, keep getting stuck on latest. 

I have a theory that this might be due to when the PR was opened, in any case I would suggest to merge to get some fresh runs, if there are more problems i ll address them in a separate PR.

joewiz:
Ok, let's try.

PR #217, opened 2019-04-10
Fix all calls to deprecated functions

wolfgangmm:
Also removes the emmet plugin, which did not work anymore.

Requires another fix to `xmldb:copy-resource` in eXist for app generation to become operational again.


duncdrum (approved):
thank you wolfgang LGTM

PR #229, opened 2019-09-03
Duncdrum patch 1

duncdrum:


adamretter (commented):
Just wondering why 4.7.1 and not 5.0.0?

duncdrum:
4.7.1 checks fo backwards compatibility with v4 series. release. is 5.0.0 and latest is 5.1.0-snapshot are moving targets. 
since we have `semver-min="4.0.0"` we should test that, if we bump to higher semver we can disable v4 checks

PR #230, opened 2019-09-03
feat(cypress): migrate int-test from e2e-core

duncdrum:


joewiz (approved):
I can see from the cypress link & info that the basic load test passes. Nice to have this integrated directly into the eXide repository so that it runs with each PR and push. Thanks @duncdrum!

PR #231, opened 2019-09-13
report exception during upload replacing old util:exception-message 

gmella:
Exception during upload are now reported properly replacing deprecated exception code (i guess).

I discovered this facing an issue uploading filename with colon inside. I will try to report this second point at right place because this is not only related to eXide...

### DIFF
--- a/modules/upload.xql
+++ b/modules/upload.xql
@@ -104,6 +104,6 @@ return
         upload:upload(xmldb:encode-uri($collection), encode-for-uri($path), $data),
         <result>
            <name>{$name}</name>
-           <error>{$util:exception-message}</error>
+           <error>{$err:description}</error>
         </result>
    )


modules/upload.xql
```diff
@@ -104,6 +104,6 @@ return
         upload:upload(xmldb:encode-uri($collection), encode-for-uri($path), $data),
```
adamretter, line 104: If you want to use `$err` then `util:catch` should be replaced with XQuery's try/catch

adamretter (changes requested):

modules/upload.xql
gmella: ok. Sorry for my confusion. It seems that this writing on master has not been ported to develop branch : https://github.com/eXist-db/eXide/blob/develop/modules/upload.xql. 
I will close this PR an try another better one. 
Hope it helps.

PR #252, opened 2020-04-20
Add handling for XSD 1.1 validation

joewiz:
The `validation:jaxv-report()` function needed for XSD 1.1 validation was added in https://github.com/eXist-db/exist/pull/1969 and released in eXist 4.3.0. 

As a result, this PR raises the minimum required version of eXist to 4.3.0 (from 4.0.0).

Since the jaxv function is only needed for XSD 1.1, we still call original jing function for other grammars.


dizzzz (approved):

adamretter (approved):

joewiz:
@dizzzz @adamretter Thank you!

PR #261, opened 2020-06-01
Capture errors without description

### LINKED ISSUE #260: [BUG] eXide hides Errors
This query currently raises a ConcurrentModificationException in eXist-db:

```xquery
<parameters>{
    let $root := <a:b xmlns:a="ans" xmlns:c="cns"/>
    for $prefix in in-scope-prefixes($root)[not(. = ('xml', 'xsi'))]
      return
          namespace-uri-for-prefix($prefix,$root) !
              <param name="{$prefix}" value="{.}"/>
}</parameters>
```

However when it is run in eXide, it just returns 0 results with no indication that an error occurred. This is misleading.

To see the exception, run the query from the REST API or Java Admin Client.

joewiz:
Closes https://github.com/eXist-db/eXide/issues/260

### DIFF
--- a/modules/session.xql
+++ b/modules/session.xql
@@ -97,12 +97,27 @@ declare function local:store-in-session($results as item()*) as element(result) 
 	can reference a result item in the session by passing parameter 'num'.
 :)
 session:create(),
-let $input := request:get-data()
+let $xqueryservlet-error := request:get-data()
 let $results := request:get-attribute("results")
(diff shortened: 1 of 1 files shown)


adamretter (approved):

adamretter:
Thanks @joewiz :-)

joewiz:
@adamretter Thank you for helping me see the issue!

PR #272, opened 2020-09-29
Add an indent parameter to the download xar endpoint

### LINKED ISSUE #271: [BUG] Download app introduces whitespace into XML documents
**Describe the bug**

From @djbpitt on [exist-open](https://exist.markmail.org/thread/5emilijstw52qxd6):

> An eXist-db app that I am developing includes an XML document with content like:
> 
> ```xml
> <form gender="m" number="sg" case="N"><stress>а</stress>ктовый</form>
> ```
> ...
> when I export the XML as part of an app using the eXide Application -> Download app menu option, the XML data line (and others like it) are changed with the addition of whitespace; they are exported as (including the newly introduced whitespace):
>
> ```xml
> <form gender="m" number="sg" case="N">
>             <stress>а</stress>ктовый</form>
> ```
>
> The introduction of whitespace during the export process damages the application; forms with added whitespace are not indexed and retrieved correctly.

**Expected behavior**

Downloading an app should not introduce whitespace into XML documents.

**To Reproduce**

See above.

**Unit Test**

n/a

**Integration Test**

An integration test would be great but I do not know how.

**Context (please always complete the following information):**

- App Version: 2.4.9

joewiz:
The new `indent=yes|no` parameter added in this PR to the deployment.xql endpoint controls serialization (i.e., pretty printing) of XML documents stored in xar files downloaded via eXide’s Application > Download app menu. By default (`indent=no`), documents will not have whitespace inserted. However, if a client sends an `indent=yes` parameter to deployment.xql, pretty printing will be enabled.

Fixes https://github.com/eXist-db/eXide/issues/271.

I manually tested and confirmed that an XML document that previously had whitespace inserted no longer does. @djbpitt tested the basic fix and confirmed it works for him too.

### DIFF
--- a/modules/deployment.xql
+++ b/modules/deployment.xql
@@ -16,7 +16,7 @@
  :  You should have received a copy of the GNU General Public License
  :  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  :)
-xquery version "3.0";
+xquery version "3.1";
 
 import module namespace apputil="http://exist-db.org/apps/eXide/apputil" at "util.xql";
 import module namespace tmpl="http://exist-db.org/xquery/template" at "tmpl.xql";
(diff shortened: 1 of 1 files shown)


joewiz:
Please disregard the failures on the branch CI - I initially pushed my branch to origin instead of my fork. I deleted this when I realized my error and pushed to my fork.

adamretter:
Two quick questions, sorry if I haven't been paying close enough attention...

1. Is the current default index=true, if so we should keep the default.
2. Presumably this applies to all XML files within a XAR, so that would mean that a XAR containing lots of data could be reformatted from what the user was expecting when it is downloaded?

djbpitt:
tl;dr: Exporting an app should not impose pretty printing. All XML should be exported as it was imported.

Slightly more detail: Adam’s two points look as if they may be mutually incompatible. The first favors keeping the current default (I guess because users might expect current behavior not to change), while the second observes that the current default risks damaging functionality. Damaging functionality is fatal; changing a default that is, ultimately, only cosmetic does not.

More detail: We pretty print only for the benefit of the human, under the assumption that it Does No Harm and makes the XML easier for the human to read. Surely we never pretty-print to introduce whitespace that is intended to be significant; if we think of it as significant, we put it there explicitly in the first place. Meanwhile, the main (perhaps only) purpose of the xar is to support distribution and installation, so if a default of pretty-printing (and changing the XML in the process) damages the data in a way that undermines the functionality of the app, making it easier for a human to read the non-functioning XML doesn't seem to convey any advantage. Shouldn't the default be that whatever the user already has installed should be exported as is (that is, in a way that is guaranteed not to impinge on functionality) when the xar is created and downloaded? That approach would guarantee no harm (if the app works, it is guaranteed to continue to work), while defaulting to pretty-printing something that is not already pretty-printed conveys no advantages with respect to installation or distribution, and risks breaking functionality.

joewiz:
@adamretter Regarding your questions:

1. eXist's conf.xml serializer defaults are set to `indent=yes`. Before this PR, eXide's "Application > Download app" command would follow this setting when stitching XML documents into xar files. This PR changes causes eXide to ignore the conf.xml setting, instead providing a parameter to control the serialization of XML documents included in xar files. The new default is `indent=no` - on the principle that, as David suggested, eXist should avoid introducing whitespace into documents.

2. It's hard to say what users expect in general, but (a) we know that many use cases require whitespace to be respected, and (b) we can clearly communicate this change in release notes. I don't foresee this change resulting in breakages. 

In sum, this is an important fix to a real bug and makes eXide a more reliable tool for users who need to generate xars from applications stored in the database. It also exposes a hook that could be used to revert to the previous behavior.

adamretter:
Okay Joe, I think I'll have to defer to you on this one, as I don't fully understand the implications

joewiz:
@eXist-db/xquery-contributors A review leading toward approval would be most appreciated.

wolfgangmm (approved):
Because I was looking into this same issue elsewhere and discussed it with Joe, I can confirm the suggested change is indeed required.

djbpitt:
Thanks, @joewiz  and @wolfgangmm, for fixing this issue!

PR #288, opened 2021-02-01
Fix build

adamretter:
The build doesn't build at the moment due to some files being moved. This fixes that.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -10,7 +10,7 @@
     <property name="templates.default.dir" value="${templates.dir}/default"/>
     <property name="xqlint.dir" value="./support/xqlint"/>
     <property name="closure.compiler.url" value="http://closure-compiler.googlecode.com/files/compiler-latest.zip"/>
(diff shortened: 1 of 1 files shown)


joewiz:
@adamretter This redirect has been handled successfully by ant for some time - for example this is the portion of my build output when rhino is downloaded:

```
get-rhino:
     [echo] Downloading rhino...
      [get] Getting: https://raw.github.com/jrburke/r.js/master/lib/rhino/js.jar
      [get] To: /Users/joe/workspace/eXide/tools/js.jar
      [get] https://raw.github.com/jrburke/r.js/master/lib/rhino/js.jar permanently moved to https://raw.githubusercontent.com/jrburke/r.js/master/lib/rhino/js.jar
      [get] Getting: https://requirejs.org/docs/release/2.1.4/r.js
      [get] To: /Users/joe/workspace/eXide/tools/r.js
```

But since it's a 301, I don't see any reason not to switch to the githubusercontent URL.

joewiz (approved):

adamretter:
Interesting, without the fix the build fails on my machine when it hits the
301

PR #306, opened 2021-05-12
Fix XSS vulnerability with open parameter

wolfgangmm:
Encode the path parameter passed in.


joewiz (approved):
I've tested and confirmed that this addresses the XSS vulnerability.

PR #309, opened 2021-06-08
Drop dependency on shared-resources

wolfgangmm:
This PR drops all dependencies on shared-resources. It also modernizes the build to facilitate future updates and replaces some (though not all) of the old components, which were depending on jquery/jquery-ui.

TODO:

- [x] replace the table grid used in the file browser. The old one shows random bugs (screen remains empty), is no longer maintained and has dependencies on outdated jquery and jquery-ui versions


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "exide",
-	"version": "2.4.13",
+	"version": "2.5.0",
```
joewiz: I think due to the removal of the app generator this might merit a 3.0.

joewiz:
This is great, Wolfgang! 

I've been testing, and other than the error reported by `npm run cypress`, the app builds and functions great. The new File > Manage window is great - I love being able to quickly filter the resources:

> Screen Shot 2021-06-08 at 5 54 54 PM

... and the advanced filters:

> Screen Shot 2021-06-08 at 5 56 52 PM

The only minor issue is that the right-most column, "Last Modified Date", is too narrow to display the date times, and AFAIK the column width can't be adjusted:

> Screen Shot 2021-06-08 at 5 57 33 PM

I think it's a great idea to use the repo to view/maintain the app's documentation. 

I'll keep using this and will report anything I see.

joewiz:
@wolfgangmm The default width of the last modified column is wonderful, and the resizable columns are working great!

I noticed one slight change: in the old file manager it was possible to rename a resource by clicking in the resource name field. I'm not able to trigger the rename mode. It was always a little fiddly, though; it was easy to accidentally trigger unintended behavior—either entering rename mode or opening the resource. Perhaps a dedicated Rename button with a modal dialog for renaming resources would be better? In any case, I just wanted to note this issue.

wolfgangmm:
@joewiz Yes, I dropped the rename feature because with the new component it is too easy to accidentally activate it. A toolbar button to click for activating the cell editor could indeed solve this.

package.json
```diff
@@ -1,30 +1,43 @@
...
-    "url": "https://github.com/eXist-db/eXide/issues"
-  },
-  "homepage": "https://github.com/eXist-db/eXide#readme",
-  "devDependencies": {
-    "cypress": "^4.12.1"
-  }
+	"name": "exide",
+	"version": "2.5.0",
+	"description": "a web-based XQuery IDE",
+	"main": "main.js",
```
joewiz, line 5: Forgive my ignorance, but given this `"main": "main.js"` object, I expected there would be a `main.js` file in the project root. I don't see one. Is this correct?

wolfgangmm: I guess this field got auto-generated. It is only relevant for library packages distributed via npm (whereas eXide is an application). It can be ignored or removed.

line-o: Yes, it is the auto generated default. For non-npm packages I usually point it to the build file if there happens to be one.

line-o: so scripts/build.js would be an option

line-o (approved):
This looks like a massive improvement.

line-o:
Well, when I try to build the project I get following error

```
➜ npm run build

> exide@2.5.0 build
> node scripts/build.js prepare && npm run xqlint && node scripts/build.js


> exide@2.5.0 xqlint
> cd support/xqlint && node r.js -o build.js && cd ../..

/Users/jll/customers/ib/eXide/support/xqlint/r.js:2267
        if (path.existsSync(url)) {
                 ^

TypeError: path.existsSync is not a function
    at requirejsVars.nodeLoad.req.load (/Users/jll/customers/ib/eXide/support/xqlint/r.js:2267:18)
    at resume (/Users/jll/customers/ib/eXide/support/xqlint/r.js:1282:65)
    at Object.require (/Users/jll/customers/ib/eXide/support/xqlint/r.js:1481:25)
    at requirejs (/Users/jll/customers/ib/eXide/support/xqlint/r.js:1684:24)
    at /Users/jll/customers/ib/eXide/support/xqlint/r.js:12506:1
    at Object.<anonymous> (/Users/jll/customers/ib/eXide/support/xqlint/r.js:12556:2)
    at Module._compile (internal/modules/cjs/loader.js:1068:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Module.load (internal/modules/cjs/loader.js:933:32)
    at Function.Module._load (internal/modules/cjs/loader.js:774:14)
```

Any pointers or ideas on the above?

wolfgangmm:
@line-o the submodule support/xqlint it should be at 36c142b12a642a5693c744afe2a0ac69c184fc9b. Maybe you still have an earlier version checked out?

line-o:
git submodule ... that's it

line-o:
Great work ... now that I was able to follow the readme and build it ;)

I found one issue, though, with the newly added filter capability in the file manager:

The second filter - that can be AND or OR  - does not seem to taken into account. It seems that it may even reset the filter completely.
I think looking at the screenshots is easier to understand:
Screenshot 2021-06-10 at 14 10 31
Screenshot 2021-06-10 at 14 10 14

line-o:
I would suggest to remove the second filter for now (hoping that it is easy to accomplish)

line-o:
The rename feature was a nuisance often times.

package.json
line-o: I tend to agree

src/commands.js
```diff
@@ -214,14 +214,14 @@ eXide.edit.commands = (function () {
...
         		    });
-                    commands.addCommand({
-            	    	name: "findModule",
-        		    	bindKey: bindKey(bindings.findModule),
-        		    	exec: function(editor) {
-                            var doc = parent.getActiveDocument();
-        		    		eXide.find.Modules.select(doc.syntax);
-        		    	}
-        		    });
+                    // commands.addCommand({
```
line-o, line 217: I would remove dead code

src/eXide.js
```diff
@@ -160,12 +156,12 @@ eXide.app = (function(util) {
 				app.saveState();
 			});
             
-            eXide.find.Modules.addEventListener("open", null, function (module) {
-                app.findDocument(module.at);
-            });
-            eXide.find.Modules.addEventListener("import", null, function (module) {
-                editor.exec("importModule", module.prefix, module.uri, module.at);
-            });
+            // eXide.find.Modules.addEventListener("open", null, function (module) {
```
line-o, line 159: same here - or do we need to fix it?

wolfgangmm: We may want to re-implement the feature sometimes, so I'll rather leave it commented out.

wolfgangmm:
The rename feature is back. You now need to click the 'rename' button in the toolbar first to switch into edit mode. This avoids any conflicts with mouse and keyboard navigation.

src/eXide.js
line-o: Well, then it is still in version control. It may never come back and would then stay there forever.
Would you be against me removing it?

line-o:
OK, everything is green. The last thing that keeps me from merging it is the discussion about dead code removal.
We could postpone this, though.
So if $someone merges this I am OK with it.

line-o:
Oh and the version discussion ...  I am in team 3.0.0

joewiz:
@wolfgangmm Thank you! The rename function works well. I just noticed one issue with it: if the user selects a cell from a field other than Name and then selects the Rename button, the contents of that cell become available for editing—instead of the contents of the Name field. Users might misunderstand that they can use this tool to change the permissions, user, group, or date of resources. Here's a screenshot showing the date field being editable:

> Screen Shot 2021-06-14 at 11 42 51 AM

If a user changes the value of one of these non-Name fields and hits return to submit the rename request, a "Rename error" dialog is validly shown: "Cannot move resource to itself '/db/test2.xlsx'. XMLDB exception caught: Cannot move resource to itself '/db/test2.xlsx'." The accompanying request URL: http://localhost:8080/exist/apps/eXide/modules/collections.xql?target=test2.xlsx&rename=test2.xlsx&root=%2Fdb%2F. Screenshot:

> Screen Shot 2021-06-14 at 11 52 26 AM

Ideally, when the Rename action is triggered, only the Name field would be made editable, regardless of which cell is selected in the row. 

Do you have a sense of whether this would be possible in this PR, or would you prefer to defer this to a future PR? I could open an issue, and we could list it as a known issue in the release.

@line-o Thanks for your fixes allowing CI checks to work with this PR!

line-o:
@wolfgangmm thanks for the hard work. This looks very well rounded now. Will merge after tests have run.

joewiz (commented):
I've tested the merged PR and can confirm that the Rename button only has effect if you've selected a resource's Name field. Thank you for everything in this PR, @wolfgangmm!

PR #314, opened 2021-06-20
Fix Slick.Grid error

wolfgangmm:
Remove the annoying, though not fatal error about Slick.Grid not being defined


line-o (approved):
Looks good, waiting for CI

PR #327, opened 2021-07-07
Add serialization prefs when opening, downloading XML documents from database

joewiz:
Before this PR eXide hard-coded serialization parameters—`indent=yes expand-xincludes=no` when opening and downloading XML documents from the database and `indent=no expand-xincludes=no` when generating and downloading packages—and users had no means to change this behavior. Numerous times the indentation behavior of eXide caused confusion when users opened documents or saved documents and observed whitespace introduced into their documents. See these posts to exist-open:

- [eXide expands XIncludes when opening XML files (2012)](https://markmail.org/message/slp5yzin6tj6y6pa) 
- [How to prevent indenting with eXide / eXist ? (2017)](https://markmail.org/message/f2fallfhh7sb6hu7)
- [how can indentation of element-only nodes be suppressed? (2020)](https://markmail.org/message/tzy6rknvqu5rpsbg) 

The only solution was to [suggest editing eXide's sources to change the hardcoded default](https://markmail.org/message/tzy6rknvqu5rpsbg) or to use a different editor.

This PR adds preferences for these two serialization parameters to the eXide preferences window:

> Screen Shot 2021-07-07 at 5 47 00 PM

The preferences default to the longstanding eXide defaults for opening and downloading individual documents: `indent=yes expand-xincludes=no`. The preferences also now govern serialization when generating and downloading packages via Application > Download app, and thus the default here is changed from a hardcoded `indent=no` to follow the eXide Preference. 

The `load.xql` endpoint now supports `indent` and `expand-xincludes` parameters, just as `deployment.xql` had already done.

Alternatives I considered:

- Showing these preferences in the editor window. I opted to place these options in Preferences since in the main editor window, they would take up valuable space; there's no good place for this in the UI; the "indent on load" checkbox could cause confusion with the "indent" checkbox in the results pane; and we've never had the ability to adjust this before, so it's doubtful we'd want/need to toggle this all of the time. Let's give this a try and see if we really need to toggle this on .
- Providing distinct preferences for (a) opening/downloading individual documents vs. (b) generating & downloading packages. I opted for a single set of preferences for simplicity's sake; it seems to me that if you like indentation, you'd want it everywhere, but if you don't want indentation, you'll disable it and want to keep it off.
- Setting the defaults via eXide's `configuration.xml` file instead of `src/preferences.js` - but none of the other preferences are set there, and I am not capable of the javascript needed to get the `preferences.js` file to read eXide's prefs from the `configuration.xml` file.

Previous work in this area includes: 

- https://github.com/eXist-db/eXide/pull/158
- https://github.com/eXist-db/eXide/pull/272

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -121,6 +121,18 @@ Drag a file on the editor to open its contents in a new tab.
 Use File > Manage > Upload Files > Upload Directory to upload
 entire directories and preserve their structure. Or drag and drop onto the Upload Files pane.
 
+## Run XQSuite tests
+
+Execute [XQSuite test suites](https://exist-db.org/exist/apps/doc/xqsuite) embedded in XQuery library modules by
+saving the module to the database (e.g., as `/db/test.xqm`) and selecting "XQuery > Run as test".
+
+## Support for EXPath Packages
+
+- Download packages stored in the database via "Application > Download app"
+- Synchronize your changes made on an application collection in the database to a directory on disk via "Application > Synchronize"
+- [Live reload](#Live-Reload) of resources in an application package
+- Upload and auto-install a package via "File > Manage > Upload" with the "Auto deploy uploaded .xar packages" checkbox selected
+
 ## Options for XML documents loaded from the database
 
 When loading XML documents from the database into an editor window, you can control whether indentation is automatically applied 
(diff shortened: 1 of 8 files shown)


joewiz:
Oops, I inadvertently pushed this to the source repository instead of my fork. Will reopen from my fork.

PR #332, opened 2021-07-15
Switch to GitHub actions

joewiz:
I adapted this workflow from the latest exist.yml in @duncdrum's generator-exist project: https://github.com/eXist-db/generator-exist/blob/e455a00479f4756b59444eb3333ea6a459aa82d0/generators/app/templates/ci/exist.yml. 

I hope that a more reliable CI here can give us confidence when reviewing Dependabot and feature/fix PRs.

If this works, it shouldn't be too hard to adapt this to the other eXist-db app repos.


joewiz:
I should add that the CI run passes on my branch: https://github.com/joewiz/eXide/actions/runs/1035146503.

Looks like it won't run here until we merge this.

Issue #334, opened 2021-07-19
[BUG] Upgrading to eXide 3.1.0 does not gracefully pick up new serialization preferences


joewiz:
**Describe the bug**

#328, included in eXide 3.1.0, added preferences for the `indent` and `expand-xincludes` serialization parameters as preferences. 

When users upgrade from earlier versions of eXide, the preferences are not registered. This causes unexpected serialization when loading documents.

Also, users don't see the new preferences unless they shift-reload the page after installing eXide.

**Expected behavior**

The new version's default behavior should match that of older versions, and new preferences should kick in gracefully. When eXide detects an old version is making a request, it should trigger a page reload to fetch the latest javascript.

**To Reproduce**

I'd like to add cypress/mocha tests for this, but until then my steps to reproduce are:

1. Perform a fresh install of eXist 5.3.0, which deploys eXide 3.0.0.
2. Open eXide 3.0.0, reload. Notice via Developer Tools > Application > Storage > Local Storage > http://localhost:8080 that preferences are stored in local storage as `eXide.preferences`.
3. Install eXide 3.1.0, reload. Notice preferences are not updated.
4. Go to Edit > Preferences > Reset Defaults. Notice preferences are updated.

**Context (please always complete the following information):**
 - OS: 11.4
 - eXist-db Version: 5.3.0
 - Java Version: 1.8
-  App Version: 3.1.0

**Additional context**
- How is eXist-db installed? DMG
- Any custom changes in e.g. `conf.xml`? n/a

PR #335, opened 2021-07-19
Fix 3.1.0 upgrade problems, restore and enhance serialization prefs

### LINKED ISSUE #333: [BUG] Restore dbutil module, formerly supplied by shared-resources
**Describe the bug**

Loading eXide 3.0.0 on a stock eXist 5.3.0 system produces a console error when eXide calls this URL:

```
http://localhost:8080/exist/apps/eXide/modules/deployment.xql?info=__new__1
```

The error:

```xml
<exception>
    <path>/db/apps/eXide/modules/deployment.xql</path>
    <message>err:XQST0059 error found while loading module dbutil: failed to load module:
        'http://exist-db.org/xquery/dbutil' from: 'http://exist-db.org/xquery/dbutil', location:
        'http://exist-db.org/xquery/dbutil'. Source not found. </message>
</exception>
```

The cause is that #309 removed the declared dependency on shared-resources but left a reference to it in `deployment.xql` and `search.xql`.

Until this module is released in a standalone package, the shortest path to a solution may be to embed the library module in eXide.

joewiz:
Fixes issues introduced in the eXide 3.1.0 release by #328.

Adds fine-grained serialization preferences:

> Screen Shot 2021-11-21 at 4 37 10 PM

Closes #333 and #334.

### DIFF
--- a/index.html.tmpl
+++ b/index.html.tmpl
@@ -646,7 +646,7 @@
                                     <label for="font-size">Font Size:</label>
                                 </li>
                                 <li>
-                                    <input type="checkbox" name="print-margin" checked="checked"/>
+                                    <input type="checkbox" name="print-margin"/>
                                     <label for="print-margin">Show Print Margin:</label>
                                 </li>
                                 <li>
--- a/index.html.tmpl
+++ b/index.html.tmpl
@@ -454,7 +454,8 @@
                           to the selected collection. Maximum file size is 100MB.</p>
                                 <form class="file_upload" action="modules/upload.xql" method="POST" enctype="multipart/form-data">
                                     <div class="options">
-                                        <input type="checkbox" name="deploy"/> Auto deploy uploaded .xar packages
+                                        <input type="checkbox" name="deploy" id="auto-deploy"/> 
+                                        <label for="auto-deploy">Auto deploy uploaded .xar packages</label>
                                     </div>
                                     <div class="file_upload_drop">Drop files here ...</div>
                                     <div class="file_upload_body">
@@ -581,6 +582,7 @@
                 <div id="preferences-dialog">
                     <form>
                         <fieldset>
+                            <legend>Editor</legend>
                             <ol>
                                 <li>
                                     <select id="theme" name="theme">
@@ -662,7 +664,7 @@
                                         <option value="4">4 spaces</option>
                                         <option value="8">8 spaces</option>
                                     </select>
-                                    <label for="indent-size">indent size:</label>
+                                    <label for="indent-size">Indent size:</label>
                                 </li>
                                 <li>
                                     <select name="soft-wrap">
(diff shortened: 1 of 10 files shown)


joewiz:
@wolfgangmm Thanks for your help on today's Community Call! I tried integrating your suggestions—in https://github.com/eXist-db/eXide/pull/335/commits/ced6868a8291939f13966bf76094f0ac98429115. But, as my comment in preferences.js line 40 states, the addition of `eXide.app.version()` raised this error in the console:

> Uncaught TypeError: Cannot read property 'version' of undefined

This is strange, because the same expression can be found elsewhere in preferences.js. Do you have any suggestions for dynamically registering the version?

joewiz:
Note that once #374 is merged, the feature commit in #328 will need to be re-added to this branch so that work on the fix started here can proceed.

src/preferences.js
```diff
@@ -173,7 +174,7 @@ eXide.util.Preferences = (function () {
     Constr.prototype.read = function() {
         var sameVersion = false;
         if (localStorage["eXide.preferences"]) {
-            this.preferences = JSON.parse(localStorage.getItem("eXide.preferences"));
+            this.preferences = Object.assign(defaultPreferences, JSON.parse(localStorage.getItem("eXide.preferences")));
```
wolfgangmm: This line should be changed as follows or the defaultPreferences would be overwritten:

```javascript
const loaded = JSON.parse(localStorage.getItem("eXide.preferences"));
this.preferences = Object.assign({}, defaultPreferences, loaded);
```

wolfgangmm (changes requested):
See my inline comment.

src/preferences.js
wolfgangmm: I tested above change on my local eXide install and it works as expected.

index.html.tmpl
```diff
@@ -683,6 +685,45 @@
...
+                        <fieldset>
+                            <legend>Serialization</legend>
+                            <p>Control (a) whether indentation is applied to XML documents or not
+                                and (b) whether XInclude elements are expanded, when opening or
+                                downloading resources and EXPath packages.</p>
+                            <h3>Opening files (via File > Open)</h3>
+                            <ol>
+                                <li>
+                                    <label for="indent-on-open">Indent:</label>
+                                    <input type="checkbox" name="indent-on-load" id="indent-on-open"/>
```
wolfgangmm: Above needs to be changed to `name="indent-on-open"` for the setting to be picked up correctly.

src/preferences.js
```diff
@@ -167,8 +186,8 @@ eXide.util.Preferences = (function () {
     Constr.prototype.read = function() {
         var sameVersion = false;
         if (localStorage["eXide.preferences"]) {
-            this.preferences = JSON.parse(localStorage.getItem("eXide.preferences"));
-            sameVersion = (this.preferences.version === eXide.app.version());
+            const loaded = JSON.parse(localStorage.getItem("eXide.preferences"));
+            this.preferences = Object.assign({}, defaultPreferences, loaded);
         }
 
         this.preferences.version = eXide.app.version();
```
wolfgangmm, line 201: Add `this.updateForm();` after `this.applyPreferences();` for the UI form to be updated.

index.html.tmpl
joewiz: @wolfgangmm Ugh, thank you for catching this! I thought I'd caught them all.

src/preferences.js
joewiz: @wolfgangmm `this.applyPreferences();` appears twice in preferences.js: first in `Constr.prototype.updatePreferences` (line 145) and second in `Constr.prototype.read` (line 195). Should I apply your suggested change to both instances?

wolfgangmm: @joewiz it should only be applied to `Constr.prototype.read`.

index.html.tmpl
joewiz: Fixed in latest force-pushed commits.

src/preferences.js
joewiz: @wolfgangmm Thank you! I've applied this change to the latest force-pushed commits. It works perfectly in my testing. Prefs are picked up without a shift-reload; a simple reload when uploading does the trick.

wolfgangmm (approved):

joewiz:
@wolfgangmm Many thanks!

Issue #322, opened 2021-07-05
[BUG] Cannot open document: "Open" button does not work


dizzzz:
Scenario NOK:
- select document
- click "Open" (red)
- 404

Scenario OK:
- double click document (blue)

![image](https://user-images.githubusercontent.com/1700062/124508823-e6cc0b00-ddd0-11eb-8209-670dde3b5dab.png)

joewiz:
I can confirm this bug with eXist 5.3.0 and eXide 3.0.0 in Chrome 91.0.4472.114 on macOS 11.4. You have to use the "open" button rather than double-clicking on a resource's row.

eXide will trigger an HTTP GET request to http://localhost:8080/exist/apps/eXide/modules/load.xql?path=undefined, returning a 404.

The expected URL is http://localhost:8080/exist/apps/eXide/modules/load.xql?path=/db/apps/demo/build.xml.

I think the relevant code is somewhere here: https://github.com/eXist-db/eXide/blob/develop/src/eXide.js#L273-L327. The code was recently touched in this commit: https://github.com/eXist-db/eXide/commit/c4821f0dae7c526f7339e6000617f5329bc11f92.

TeddyBoomer:
Hi, I would like to point out a  hint(?) about this problem but I'm rather newbie on js:
in [https://github.com/eXist-db/eXide/blob/develop/src/resources.js#L155](https://github.com/eXist-db/eXide/blob/develop/src/resources.js#L155): 
there is a call in a .api attribute while it's undefined above at L147-152; and then the button can't grab the right data, while onRowDoubleClicked is ok just after at L163.

(12/09/2021)
Hi, ok -1 for me: I got into ag-grid some times ago, for the .api.
Coming back to joewitz  https://github.com/eXist-db/eXide/blob/develop/src/eXide.js#L273-L327. 
I think the "open" definition of the button is pretty strange: openSelectedDocument is a 2 params function
for a click (no params), maybe:
`"open": function() { openSelectedDocument(null, false);}`
would be better at L278 (still don't know why it would work before with only one param)

PR #371, opened 2021-10-09
fix eXide #322. seems to work fine after compile and load in the app-…

TeddyBoomer:
…manager

1st PR for me: 1 comment this summer on eXide #322 and a mail to joewiz about that.
please check that it's really ok…

### DIFF
--- a/src/eXide.js
+++ b/src/eXide.js
@@ -274,8 +274,8 @@ eXide.app = (function(util) {
 			dbBrowser.reload(["reload"], "open");
 			$("#open-dialog").dialog("option", "title", "Open Document");
 			$("#open-dialog").dialog("option", "buttons", { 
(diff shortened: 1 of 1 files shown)


joewiz (changes requested):
@TeddyBoomer Your PR is great! Thank you so much! I just noticed one slight issue with the updated behavior: after opening a document via the Open button, the dialog remains  open. I'd expect it to close automatically. In my tinkering, I noticed that if you change the 2nd parameter of openSelectedDocument to `true`, the dialog will close automatically. Would you make this change to the PR? If it's inconvenient, I can apply the change. Either way is fine with me.

TeddyBoomer:
Hi Joewiz, I've toggled the 2d parameter; it appears in the PR.
Best regards

joewiz (approved):
@TeddyBoomer Perfect. Thanks so much for your contribution!

PR #420, opened 2022-02-06
Fix snippets for XQuery 3.1

joewiz:
The snippet with the XQuery version declaration contained a syntax error. 

The snippet with the HTML5 serialization is updated to conform to the XSLT and XQuery Serialization 3.1 spec, at https://www.w3.org/TR/xslt-xquery-serialization-31/#HTML_VERSION. While eXist is forgiving about `output:method="html5"`, this formulation will throw an error in Saxon and BaseX.

### DIFF
--- a/templates/xquery.snippets
+++ b/templates/xquery.snippets
@@ -65,7 +65,8 @@ snippet json
 	declare option output:media-type "application/json";
 	
 snippet html5
-	declare option output:method "html5";
+	declare option output:method "html";
+	declare option output:html-version "5.0";
 	declare option output:media-type "text/html";
 
 snippet html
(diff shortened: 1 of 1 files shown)


joewiz:
@wolfgangmm I'm curious to know: Is the [snippets.xml](https://github.com/eXist-db/eXide/blob/develop/templates/snippets.xml) used to generate the source of [xquery.snippets](https://github.com/eXist-db/eXide/blob/develop/templates/xquery.snippets)? In testing this PR, my edits - which were only to xquery.snippets - had the desired effect in eXide. But I just stumbled upon snippets.xml and realized that perhaps I should be editing that and somehow generating the xquery.snippets file from it. I'm happy to adjust the PR, but I would appreciate any guidance on how to generate the xquery.snippets file. I wasn't able to find anything in source about this. Alternatively, we could delete the snippets.xml file. Thanks!

line-o (approved):

joewiz:
@line-o Thank you!

PR #433, opened 2022-02-21
Ensure errors in module inspection don’t halt docs

joewiz:
In certain cases, an installed module is incompatible with eXist. For example, the current release of the EXPath Crypto module is [incompatible](https://github.com/eXist-db/expath-crypto-module/issues/64) with the current release of eXist. In this situation, passing this module’s URI to inspect:inspect-module-uri will throw an error. Without wrapping this call in a try-catch block, this error will cause requests to eXide’s docs.xql endpoint to return a server error. Wrapping them in a try-catch block allows the call to docs.xql to complete. Documentation for the failing module (crypto, in this scenario) won’t be available, obviously, but documentation for all other modules will be available.


duncdrum (approved):

joewiz:
@line-o Thank you!

PR #468, opened 2022-05-19
Fix for login-dialog to focus into user box on load.

### LINKED ISSUE #109: Login dialog should focus on User input field with File > Manage
When selecting File > Manage, the Login dialog does not put focus in the User input field. Typing in this state results in text being entered into the editor rather than the User input field, and there's no keyboard shortcut for cycling focus into the User field. Note that logging in via the Login button does correctly move focus to the User field.

cmccrorie:
Closes https://github.com/eXist-db/eXide/issues/109.

Description: This Pull request is to fix the issue with the login Dialog so that it now focuses into the user box when the user is not logged in.

Desired behaviour: When someone is not logged in, and either clicks on `File > manage` and/or `File > save (and save-as)`, when the login-dialog form appears, it automatically focuses into the user text box rather than the editor still.  

### DIFF
--- a/src/menu.js
+++ b/src/menu.js
@@ -113,6 +113,9 @@ eXide.util.Menubar = (function() {
             if ($this.editor) {
                 $this.editor.focus();
             }
+            if($('#login-dialog').is(':visible')) {
+                $("#login-dialog input:first").focus();
+            }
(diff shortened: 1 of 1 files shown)


joewiz (approved):
I can confirm the fix works for me.

PR #469, opened 2022-05-23
Save as uses current file's collection closes #104

### LINKED ISSUE #104: Save as should use current file's collection
Upon launch eXide's File > Save As dialog uses `/db` as the collection. Instead, it should use the currently opened document's collection. Looking at the source, it appears the code is trying to do this, but for me in Chrome, it starts with `/db`. 

  https://github.com/wolfgangmm/eXide/blob/develop/src/eXide.js#L430-L453

marmoure:
Fixes #104

 eXide's File > Save As dialog now uses current file collection as the collection.


joewiz (approved):
I can confirm that this fixes the issue.

PR #471, opened 2022-05-25
modify comment logic and add more cases

### LINKED ISSUE #112: Autocomplete fails to notice that a comment block is already closed
After closing an XQuery comment, pressing return causes a newline with a space + colon:

```
(: this is a comment :)
 : 
```

It seems like eXide's autocomplete function is trying to help keep comment blocks prepended with `:`, but it fails to detect that the comment block is already closed.

marmoure:
fixes #112
After closing an XQuery comment, pressing return causes a newline with a space + colon:
```
(: this is a comment :)
 : 
```
We adjusted the logic that decides when a colon should spawn to prevent the case above.
And here are the cases of when a colon should spawn:
* the cursor is in between the comment brackets `(:  |  :)`
* the line start with a `(:` and there is no closing bracket 
* you are in the middle of a comment and the line starts with ` :` 

### DIFF
--- a/src/ace-modes.js
+++ b/src/ace-modes.js
@@ -197,7 +197,7 @@ define("eXide/mode/behaviour/xquery", function(require, exports, module) {
                 }
             }
             if (text == ":") {
-                if (line.slice(cursor.column -1 ,cursor.column)) {
+                if (line.slice(cursor.column -1 ,cursor.column) === '(') {
                     return {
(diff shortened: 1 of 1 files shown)


marmoure:
@joewiz i added the missing case and the comment logic seems okayish

joewiz (approved):
Works perfectly and fixes the issue I reported.

PR #475, opened 2022-05-30
make auto pairing optional for quotes and parens

### LINKED ISSUE #117: Add option to disable auto-pair of quotes, parens, etc.
I've never gotten used to eXide's auto-pair function.  This is where you type a single `"` and get two - `""` - with your cursor placed between them.  The same behavior affects `'`, `(`, and `{`.  Often, when I am trying to insert a single instance of these characters, the editor inserts two.  Then I delete one, but the editor deletes both.  This can be maddening and leads to many extraneous key presses.  I imagine some developers are used to this behavior from other code editors, but for me - and I'd bet - many new users, typing one thing and having something else happen is disconcerting and leads to wasted keypresses. 

Would it be possible to add a preference disabling this?

Caveat: I would hope that this preference wouldn't disable the autocompletion of elements, e.g., `<x>as soon as I type </`immediately becomes `<x>as soon as I type </x>`.  This generally works well - not perfect, but on balance is a time saver.  Also valuable is the snippet behavior, where I can type a code like `html`, tap control-space, and `html` is replaced by some XQuery code.

marmoure:
this PR introduce a new feature that allows to toggle auto pairing for quotas and parens 
Fixes #117 
image

### DIFF
--- a/controller.xq
+++ b/controller.xq
@@ -63,9 +63,9 @@ declare function local:fallback-login($domain as xs:string, $maxAge as xs:dayTim
 
 declare function local:user-allowed() {
     (
-        request:get-attribute("org.exist.login.user") and
-        request:get-attribute("org.exist.login.user") != "guest"
-    ) or config:get-configuration()/restrictions/@guest = "yes"
+        session:get-attribute("org.exist.login.user") and 
+        session:get-attribute("org.exist.login.user") ne "guest" 
+        ) or config:get-configuration()/restrictions/@guest eq "yes"
 };
 
 declare function local:query-execution-allowed() {
--- a/index.html.tmpl
+++ b/index.html.tmpl
@@ -682,7 +682,7 @@
                                 </li>
                                 <li>
                                     <input type="checkbox" name="auto-pair"/>
-                                    <label for="auto-pair">auto pair:</label>
+                                    <label for="auto-pair">Auto-pair characters:</label>
                                 </li>
                                 <li>
                                     <input type="checkbox" name="emmet"/>
--- a/index.html.tmpl
+++ b/index.html.tmpl
@@ -158,6 +158,10 @@
                                     <a href="#" id="menu-edit-preferences" data-command="preferences">
                                         <span class="fa fa-cogs"/>Preferences</a>
                                 </li>
+                                <li>
+                                    <a href="#" id="menu-toggle-autoPair" data-command="toggleAutoPair">
+                                        <span class="fa fa-check"/>autoPair</a>
+                                </li>
                             </ul>
                         </li>
                         <li>
--- a/index.html.tmpl
+++ b/index.html.tmpl
@@ -158,10 +158,6 @@
                                     <a href="#" id="menu-edit-preferences" data-command="preferences">
                                         <span class="fa fa-cogs"/>Preferences</a>
                                 </li>
-                                <li>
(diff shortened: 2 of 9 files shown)


marmoure:
@cmccrorie  this needs review and testing

cmccrorie:
> @cmccrorie this needs review and testing

@marmoure I have taken a look at this and the option to enable and disable this with in `Edit > autopair` works. I have had no issue when testing this.

joewiz:
@marmoure In my testing, toggling the new "Edit > autoPair" command does toggle the auto-pairing of single quotes, double quotes, and parentheses: `'` `"` `(`. This is great! I have a couple of questions:

1. Reloading the page resets the current selection for the toggle, but the setting should be persistent across reloads of the application, and it should live in "Edit > Preferences" as a checkbox. Could you please move this setting into "Edit > Preferences" as a checkbox with the label "Auto-pair characters"? 
2. With auto-pairing on, square and curly braces - `[` `{` - are not auto-paired, but should be. Would it be easy/quick to extend the auto-pairing behavior to these characters?

Reminder to self: add a note about this new option to https://github.com/eXist-db/eXide/blob/develop/docs/docs.md (perhaps under "code completion"?) before releasing.

marmoure:
thanks @joewiz sounds good
i ll take a look, for the braces i didn't change the original behavior when auto pair is on,would you like to do so?

joewiz:
@marmoure Interesting! In my original note, I mentioned that square brackets were being auto-paired, but perhaps I was mistaken or the behavior changed in the meantime. (I see square brackets are auto-paired in the demo at https://ace.c9.io/.) 

Yes, if it's not too much trouble, I think that in the XQuery context it's natural to have auto-pairing on these two characters - `[` and `{`. 

But if it looks complex, the PR is acceptable without extending auto-pairing to these characters.

Thanks!

joewiz:
(I see other ace users have requested more configurable behavior regarding auto-pairing... https://github.com/ajaxorg/ace/issues/4199.)

marmoure:
@joewiz we can have that as a feature if needed 
we can configure the wanted behavior in eXide

marmoure:
@joewiz  the autoPair is now in `EDIT -> Preferences` and it's persistent 
doing it this way makes it enabled globally for all documents as per the previous approach  it was document independent 
image

joewiz:
I can confirm that the auto-pair preference works as expected. It changes operations immediately and persists across reloads of eXide.

Could you fix the preference label? It should be "Auto-pair characters" rather than "auto pair".

joewiz:
@marmoure GitHub is reporting a conflict in the PR.

marmoure:
@joewiz  can you give this one, a final spin and see the output

joewiz:
@marmoure I see that when I enter `[` or `{` the corresponding closing character is added, with the cursor going between the two characters, as expected. But when I hit the delete key to delete the opening character, both the opening and closing character should be deleted. This is the case with `'`, `"`, and `(` pairs. Would it be possible to ensure the same behavior in the case of `[` and `{`?

marmoure:
@joewiz  i ll try to copy the behavior from  the double quotes, it's a nice suggestion

joewiz (approved):
Works perfectly!

joewiz:
@marmoure Thank you!

PR #484, opened 2022-06-17
Add download button to DB Manager view 

### LINKED ISSUE #111: Add download button to DB Manager view
Currently in eXide, to download a file, you must first open the file and then select File > Download. Some files are so large that opening them is slow - inefficient just to download a file. It would be very convenient if the DB Manager view (File > Manage) could add a Download button to download a selected resource. 

The ability to download collections is less important, but perhaps if a download feature were to be extended to collections (or multiple selected resources), the selected collections/resources could be downloaded as a zip file - much as the Application > Download App button does.

marmoure:
fixes #111 
 DB Manager view (File > Manage) now has a Download button to download a selected resource.
 This only applies for files at the moment, you select the file then press Download in the toolbar.
 
 image

### DIFF
--- a/src/eXide.js
+++ b/src/eXide.js
@@ -289,7 +289,7 @@ eXide.app = (function(util) {
 				$("#open-dialog").dialog("close");
 		},
 
-        saveSelectedDocument: function(doc, close) {
+        downloadSelectedDocument: function(doc, close) {
 			var resource = doc || dbBrowser.getSelection();
 			if (resource) {
 				app.download(resource.path);
--- a/src/eXide.js
+++ b/src/eXide.js
@@ -480,7 +480,7 @@ eXide.app = (function(util) {
 				return;
 			}
             var path = path || doc.getPath()
-            window.location.href = "modules/load.xql?download=true&path=" + encodeURIComponent(path) + "&indent=" + indentOnDownload + "&expand-xincludes=" + expandXIncludesOnDownload + "&omit-xml-decl=" + omitXMLDeclatarionOnDownload;
+            window.location.href = "modules/load.xq?download=true&path=" + encodeURIComponent(path) + "&indent=" + indentOnDownload + "&expand-xincludes=" + expandXIncludesOnDownload + "&omit-xml-decl=" + omitXMLDeclatarionOnDownload;
 		},
         
 		runQuery: function(path, livePreview) {
--- a/src/eXide.js
+++ b/src/eXide.js
@@ -292,14 +292,14 @@ eXide.app = (function(util) {
         downloadSelectedDocument: function(docs, close) {
 			var resources = docs;
 			if (resources) {
-                let collectionsExist = resources.filter(res => res.isCollection).length > 0;
-                if(!collectionsExist) {
-                    resources.forEach(resource => {
-                        app.download(resource.key);
-                    })
-                }else {
-                    util.Dialog.warning("Invalid selection","The Download Selected command requires that only resources be selected (not collections). Please select a resources for download.")
-                }
+                resources.filter(res => res.isCollection).forEach(collection => {
+                    deploymentEditor.download(collection.key);
+                });
+
+                resources.filter(res => !res.isCollection).forEach(resource => {
+                    app.download(resource.key);
+                })
+
 			}else {
                 util.Dialog.warning("Invalid selection","The Download Selected command requires that resources be selected. Please select a resources for download.")
             }
--- a/src/eXide.js
+++ b/src/eXide.js
@@ -289,6 +289,15 @@ eXide.app = (function(util) {
 				$("#open-dialog").dialog("close");
 		},
 
+        saveSelectedDocument: function(doc, close) {
+			var resource = doc || dbBrowser.getSelection();
+			if (resource) {
+				app.download(resource.path);
+			}
+			if (close == undefined || close)
+				$("#open-dialog").dialog("close");
+		},
+
 		$doOpenDocument: function(resource, callback, reload) {
 			resource.path = util.normalizePath(resource.path);
             var indentOnOpen = $("#indent-on-open").is(":checked");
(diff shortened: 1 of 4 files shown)


src/eXide.js
```diff
@@ -461,16 +470,17 @@ eXide.app = (function(util) {
...
             var omitXMLDeclatarionOnDownload = $("#omit-xml-decl-on-download").is(":checked");
 			var doc = editor.getActiveDocument();
-			if (doc.getPath().match("^__new__") || !doc.isSaved()) {
+			if (!path && (doc.getPath().match("^__new__") || !doc.isSaved())) {
 				util.error("There are unsaved changes in the document. Please save it first.");
 				return;
 			}
-            window.location.href = "modules/load.xq?download=true&path=" + encodeURIComponent(doc.getPath()) + "&indent=" + indentOnDownload + "&expand-xincludes=" + expandXIncludesOnDownload + "&omit-xml-decl=" + omitXMLDeclatarionOnDownload;
+            var path = path || doc.getPath()
+            window.location.href = "modules/load.xql?download=true&path=" + encodeURIComponent(path) + "&indent=" + indentOnDownload + "&expand-xincludes=" + expandXIncludesOnDownload + "&omit-xml-decl=" + omitXMLDeclatarionOnDownload;
```
joewiz: @marmoure As of eXide [3.4.0](https://github.com/eXist-db/eXide/releases/tag/v3.4.0), the load endpoint is `load.xq`, not `load.xql`.

marmoure:
@joewiz  i updated the file name ,please take a look at PR

src/eXide.js
```diff
@@ -289,6 +289,15 @@ eXide.app = (function(util) {
 				$("#open-dialog").dialog("close");
 		},
 
+        saveSelectedDocument: function(doc, close) {
```
joewiz: @marmoure Would this function be better named `downloadSelectedDocument` or `downloadSelectedResource`? Many other functions in this file already use "save" to mean "save a resource to the database," so I worry that the name "saveSelectedDocument" could lead to some confusion.

joewiz:
@marmoure Thanks for fixing the call to `load.xq`! It works well for downloading individual resources. Since this PR doesn't allow download of multiple selected resources or collections (i.e., the operation silently fails), could you either (a) make the download button inactive when multiple resources or collections are selected, or (b) display a notice when the user selects the download button in such a scenario?

marmoure:
hi @joewiz  i added a fail message when eXide fail to recognize your selection 
and i ll move my comment from the original issue to here 
"Hi @joewiz seems like getting selected resources in eXide in the current state doesn't support selecting collections or multiple resources, you can only select one resource"

joewiz:
Thanks! Could you update the text of the warning from:

> **Error in selection** eXide didn't recognize your selection please select a single resource first. current state doesn't support selecting collections or multiple resources

to:

> **Invalid selection**: The Download Selected command requires that only one resource be selected (not multiple resources or collections). Please select a single resource for download.

(In my testing, the copy/paste and cut/paste commands can work on multiple resources. But the commands are somewhat inconsistent in their handling of multiple selections. Your warning dialog is very nice and user-friendly!)

marmoure:
Hi @joewiz  i changed the warning message and added few tweaks 😉

joewiz:
@marmoure Works beautifully! Could you change the warning from:

> The Download Selected command requires that only resources be selected (not collections). Please select a resources for download.

to:

> The Download Selected command only works on resources, not collections. Please select resources for download.

joewiz:
@marmoure Hmm... If we extend the `deploy:download()` function in [modules/deployment.xq](https://github.com/eXist-db/eXide/blob/develop/modules/deployment.xq#L588-L621) to not just create .xar packages but also create .zip files for collections, as follows:

```xquery
declare function deploy:download($collection as xs:string, $expathConf as element()?, $expand-xincludes as xs:boolean, $indent as xs:boolean, $omit-xml-declaration as xs:boolean) {
    let $name := 
        if ($expathConf) then
            concat($expathConf/@abbrev, "-", $expathConf/@version, ".xar")
        else
            replace($collection, "^.+/([^/]+)$", "$1") || ".zip"
    let $entries :=
        (: compression:zip uses default serialization parameters, so we'll construct entries manually :)
        dbutil:scan(xs:anyURI($collection), function($coll as xs:anyURI, $res as xs:anyURI?) {
            (: compression:zip doesn't seem to store empty collections, so we'll scan for only resources :)
            if (exists($res)) then
                let $relative-path := substring-after($res, $collection || "/")
                return
                    if (util:binary-doc-available($res)) then
                        <entry type="uri" name="{$relative-path}">{$res}</entry>
                    else
                        <entry type="xml" name="{$relative-path}">{
                            (: workaround until https://github.com/eXist-db/exist/issues/2394 is resolved :)
                            util:declare-option(
                                "exist:serialize", 
                                "expand-xincludes=" 
                                || (if ($expand-xincludes) then "yes" else "no")
                                || " indent=" 
                                || (if ($indent) then "yes" else "no")
                                || " omit-xml-declaration=" 
                                || (if ($omit-xml-declaration) then "yes" else "no")
                            ),
                            doc($res)
                        }</entry>
            else
                ()
        })
    let $archive := compression:zip($entries, true())
    return (
        response:set-header("Content-Disposition", concat("attachment; filename=", $name)),
        response:stream-binary($archive, "application/zip", $name)
    )
};
```

... then this PR could be extended to download collections (like `/db/test`) by calling:

http://localhost:8080/exist/apps/eXide/modules/deployment.xq?download=true&collection=%2Fdb%2Ftest&indent=false&expand-xincludes=false&omit-xml-decl=true

The `collection` param should contain the full, URL-encoded path to the desired collection. The `indent`, `expand-xincludes`, and `omit-xml-decl` parameters should be read from preferences, as [other similar commands](https://github.com/eXist-db/eXide/blob/develop/src/deployment.js#L214-L219) do.

Do you think that's feasible to incorporate into this PR?

marmoure:
@joewiz can you please go ahead and give this one a test

joewiz:
@marmoure Works beautifully - even with mixed selections of resources AND collections. Thank you!

joewiz (approved):

PR #485, opened 2022-06-17
Re Enable login With configuration.xml set to guest=no

### LINKED ISSUE #164: With configuration.xml set to guest=no, login is impossible
In my testing of @line-o's https://github.com/eXist-db/eXide/pull/142, setting `guest=no` prevents users from even getting to the page where they would log in.  I propose reverting this PR and restoring the ability of users to access the log in page so that they can authenticate.

marmoure:
Fixes #164 
The `local:user-allowed()` function now checks against `session` attributes instead of `request` attributes.
Implementing @rvdb fix stated here https://github.com/eXist-db/eXide/issues/164#issuecomment-342827531
This now re-enable login for guest user when configuration.xml set to guest=no 

### DIFF
--- a/controller.xq
+++ b/controller.xq
@@ -64,8 +64,8 @@ declare function local:fallback-login($domain as xs:string, $maxAge as xs:dayTim
 declare function local:user-allowed() {
     (
         session:get-attribute("org.exist.login.user") and 
-        session:get-attribute("org.exist.login.user") != "guest" 
-        ) or config:get-configuration()/restrictions/@guest = "yes"
+        session:get-attribute("org.exist.login.user") ne "guest" 
+        ) or config:get-configuration()/restrictions/@guest eq "yes"
 };
 
 declare function local:query-execution-allowed() {
--- a/controller.xq
+++ b/controller.xq
@@ -63,9 +63,9 @@ declare function local:fallback-login($domain as xs:string, $maxAge as xs:dayTim
 
 declare function local:user-allowed() {
     (
-        request:get-attribute("org.exist.login.user") and
-        request:get-attribute("org.exist.login.user") != "guest"
-    ) or config:get-configuration()/restrictions/@guest = "yes"
+        session:get-attribute("org.exist.login.user") and 
+        session:get-attribute("org.exist.login.user") != "guest" 
+        ) or config:get-configuration()/restrictions/@guest = "yes"
 };
 
 declare function local:query-execution-allowed() {


joewiz:
I can confirm that with this PR, when `guest="no"` is set in eXide's `configuration.xml` file, guest users can log in via Dashboard and then access eXide. Before this PR, even after authenticating as a non-guest user in Dashboard, eXide would refuse to let authenticated users into the application. 

Looking back over the history of this issue, I don't think we ever learned the answer to @rvdb's question about [the original PR](https://github.com/eXist-db/eXide/pull/142) that instituted this severe approach toward authentication of guest users. As he wrote in https://github.com/eXist-db/eXide/pull/142#discussion_r149674898: 

> This makes the eXide login screen unavailable to unauthenticated users, thus making it impossible to log in via eXide. Is this the purpose of this check?

I'd still be interested to know why eXide shouldn't display its own login dialog! 

But, in the absence of an answer... I think this PR does improve the situation.

Thanks to @rvdb for the fix incorporated here and to @dscheuneman for confirming the issue and fix.

Note to self: Be sure to update https://github.com/eXist-db/eXide/blob/develop/docs/docs.md#security with the suggested language in https://github.com/eXist-db/eXide/issues/164#issuecomment-301091811 before releasing the version of eXide incorporating this fix.

joewiz (approved):

controller.xq
```diff
@@ -63,9 +63,9 @@ declare function local:fallback-login($domain as xs:string, $maxAge as xs:dayTim
 
 declare function local:user-allowed() {
     (
-        request:get-attribute("org.exist.login.user") and
-        request:get-attribute("org.exist.login.user") != "guest"
-    ) or config:get-configuration()/restrictions/@guest = "yes"
+        session:get-attribute("org.exist.login.user") and 
+        session:get-attribute("org.exist.login.user") != "guest"
```
adamretter: Do we really mean `!=` to here? That operator in XQuery is probably not doing what you expect - you may have menat `ne` instead of `!=`

controller.xq
```diff
@@ -63,9 +63,9 @@ declare function local:fallback-login($domain as xs:string, $maxAge as xs:dayTim
 
 declare function local:user-allowed() {
     (
-        request:get-attribute("org.exist.login.user") and
-        request:get-attribute("org.exist.login.user") != "guest"
-    ) or config:get-configuration()/restrictions/@guest = "yes"
+        session:get-attribute("org.exist.login.user") and 
+        session:get-attribute("org.exist.login.user") != "guest" 
+        ) or config:get-configuration()/restrictions/@guest = "yes"
```
adamretter: `=` should be replaced with `eq` here.

PR #491, opened 2022-06-29
[feat] added syntax highlighting to query results

### LINKED ISSUE #147: Switch to xqlint for syntax highlighting of query results
Several attempts at fixing #108 had unintended side effects (and thus were abandoned). While there's probably a solution to that issue, it occurs to me that pretty-print.xql does just a so-so job because it essentially re-implements serialization in XQuery. Rather than continuing to expand it with other output modes, eXist already does an excellent job of serializing, and xqlint already does an excellent job of applying syntax highlighting. 

Perhaps we should switch eXide to using xqlint for syntax highlighting of query results. Atom-existdb does (see https://github.com/wolfgangmm/atom-existdb/blob/master/lib/existdb.coffee#L145-L152 and https://github.com/wolfgangmm/atom-existdb/blob/master/lib/xquery-helper.coffee#L7-L14) this successfully. atom-existdb currently assumes XML output, so doesn't currently support JSON or adaptive output, but I'm guessing JSON would be very straightforward to support, and adaptive could work if exist-db were to add native support for adaptive serialization.

The only thing we'd lose is the match highlighting of query results. And we could probably find a workaround for that - perhaps by extending adaptive serialization with some match highlighting markers.

ccheraa:
This adds syntax highlighting to query results.

Closes #147.

This open source contribution to the eXide project was commissioned by the Office of the Historian, U.S. Department of State, https://history.state.gov/.

### DIFF
--- a/resources/css/eXide.css
+++ b/resources/css/eXide.css
@@ -898,6 +898,10 @@ Sub level menu list items (undo style from Top level List Items)
 
 .item {
     min-height: 2.5em;
+    -ms-user-select: text;
+    -moz-user-select: text;
+    -webkit-user-select: text;
+    user-select: text;
 }
 
 /* Hack for internet explorer */
--- a/resources/css/eXide.css
+++ b/resources/css/eXide.css
@@ -909,12 +909,16 @@ Sub level menu list items (undo style from Top level List Items)
 }
 
 .even {
-    /*background-color: #F0F0F0; */
+    background-color: rgba(255, 255, 255, 0.02);
+}
+
+.uneven {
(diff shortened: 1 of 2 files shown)


joewiz:
@ccheraa Very nice! Just one issue upon a very quick review:

It seems that text in the results pane cannot be selected - so copy and paste of content from the results pane is not possible. Could you look into restoring the ability to select spans of text in the results pane?

joewiz (approved):
Text selection works perfectly. Syntax highlighting of results has no discernible effect on performance.

joewiz:
@ccheraa Thank you!

PR #503, opened 2022-07-20
Use separate names for CI artifacts from different builds

adamretter:


joewiz (approved):
As shown in https://github.com/eXist-db/eXide/actions/runs/2705777467, all 4 sets of per-build artifacts are now uploaded:

> Screen Shot 2022-07-20 at 11 26 44 AM

Issue #499, opened 2022-07-13
Add a button to copy a result to the clipboard


joewiz:
**Is your feature request related to a problem? Please describe.**

Often users want to copy the results of a query to the clipboard. For long results it is cumbersome to select the entire result.

**Describe the solution you'd like**

Add a button beneath each "numbered result", with the label "copy" or the clipboard emoji 📋. Selecting this button would copy the string value of the item to the clipboard. The solution could use https://clipboardjs.com/.

With #491 results are displayed with syntax highlighting. I think the clipboard should only copy the plain text, if possible.

All Output methods except Direct Output support the "numbered result" display described above. So it might not be possible to support the copy to clipboard feature for Direct Output.

**Describe alternatives you've considered**

The copy button *could* copy syntax-highlighted results (added with #491), and then users would have to use their own methods to "paste as plain text" or "paste and match style", etc.

**Additional context**

n/a

marmoure:
this should be closed as #513 is merged!

joewiz:
While #513 is a huge step forward and a wonderful new feature, this issue has a slightly different use case: the need to copy a single result to the clipboard. While, together with #527, a user could limit the results to a single result and use next/prev buttons to get just the one they want, and then use the copy button, that is not as simple as the feature envisioned here. So I’d like to keep it open.

joewiz:
Closing as a dupe of #97

PR #513, opened 2022-08-11
feature copy results to clipboard

marmoure:
while working with eXide and running eval on some scripts i found that copying the results was not easy at all.
first i can't select it all using ctrl-A as the focus was always on the editor pane and trying to do it with the mouse proven to be a hassle.
i decided to add a copy to clipboard button which copies all the results showing in the results pane.
image

### DIFF
--- a/index.html.tmpl
+++ b/index.html.tmpl
@@ -419,8 +419,8 @@
                                         <input type="checkbox" id="auto-expand-matches" name="auto-expand-matches" checked="checked"/>
                                         Highlight Index Matches
                                     </label>
-                                    <label id="copy-all-clipboard" title="copy the resutls into the clipboard">
-                                    <i class="fa fa-clipboard" aria-hidden="true"></i>copy to clipboard
+                                    <label id="copy-all-clipboard" title="Copy the displayed results into the clipboard">
+                                    <i class="fa fa-clipboard" aria-hidden="true"></i> Copy to clipboard
                                     </label>
                                     <img class="layout-switcher" src="resources/images/layouts_split.png" title="Switch results panel location between bottom and right"/>
                                     <span class="current"/>
--- a/index.html.tmpl
+++ b/index.html.tmpl
@@ -419,6 +419,9 @@
                                         <input type="checkbox" id="auto-expand-matches" name="auto-expand-matches" checked="checked"/>
                                         Highlight Index Matches
                                     </label>
+                                    <label id="copy-all-clipboard" title="copy the resutls into the clipboard">
+                                    <i class="fa fa-clipboard" aria-hidden="true"></i>copy to clipboard
+                                    </label>
                                     <img class="layout-switcher" src="resources/images/layouts_split.png" title="Switch results panel location between bottom and right"/>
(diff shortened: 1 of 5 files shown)


index.html.tmpl
```diff
@@ -419,6 +419,9 @@
                                         <input type="checkbox" id="auto-expand-matches" name="auto-expand-matches" checked="checked"/>
                                         Highlight Index Matches
                                     </label>
+                                    <label id="copy-all-clipboard" title="copy the resutls into the clipboard">
+                                    <i class="fa fa-clipboard" aria-hidden="true"></i>copy to clipboard
+                                    </label>
```
joewiz:
```suggestion
                                    <label id="copy-all-clipboard" title="Copy the displayed results into the clipboard">
                                    <i class="fa fa-clipboard" aria-hidden="true"></i> Copy to clipboard
                                    </label>
```

src/eXide.js
```diff
@@ -1551,6 +1551,16 @@ eXide.app = (function(util) {
                 util.requestFullScreen(document.getElementById("fullscreen"));
             });
             $(".results-container .layout-switcher").click(app.switchResultsPanel);
+			$('.results-container #copy-all-clipboard').click(() => {
+                let res = '';
+                document.querySelectorAll("#results-body > div > div > div > div.item").forEach(item => res += item.innerText)
+                navigator.clipboard.writeText(res).then(function() {
+                    console.log('Async: Copying to clipboard was successful!');
+                    eXide.util.message("Copyed to clipboard");
```
joewiz:
```suggestion
                    eXide.util.message("Copied results to clipboard");
```

joewiz (approved):
It works great! Just a couple of suggestions in the wording. Thanks for adding tests too! I notice these appear to be failing?

marmoure:
@joewiz yes seems like CI test for copying doesn't work in this environment.
the clipboard requires special permissions and i tried to work around that
but seems like it failed. 
can you please run the test locally

joewiz:
@marmoure I also get the same "Browser context management is not supported" error shown in the CI log when running `npm run cypress`.

marmoure:
after updating the test i m still seeing a failure for checking the notification which is there???!!
Navbar -- should display notification (failed)

marmoure:
@joewiz this unfortunately needs a lot of work to fix the broken tests
it Fixes #499 
how should we progress?

joewiz:
@marmoure Ah, I'm not sure... Is there a way to keep the tests while marking them as "pending" or "don't run"? Perhaps the situation with browser context management will change in the future, and we could just re-enable your tests.

Re: #499, this PR is indeed quite similar, but my original idea was to allow copying on a per-item basis, but yours copies the current 10 results as a whole. I don't suppose it would be hard to add a per-item copy function? I'm happy to merge this without that, but I thought I'd ask in case it helps.

marmoure:
@joewiz can you please elaborate and how that function work When you have multiple results.

joewiz (approved):
As discussed on today's call, we will merge this as is.

@marmoure Thank you!

PR #594, opened 2023-02-13
Rework controller and autorization flow

line-o:
Fixes #164

- add login page as static HTML (adapted from dashboard)
- rewrite controller to always authenticate requests
- add routes handling new flow
- add background SVG for login page

Here is a screenshot of Safari (top left), Edge (top right), Chrome (bottom left), Firefox (bottom right) showing the login page.

Screenshot 2023-02-13 at 21 35 25

PR #595, opened 2023-02-13
Rework controller and autorization flow

### LINKED ISSUE #164: With configuration.xml set to guest=no, login is impossible

line-o:
Fixes #164

- add login page as static HTML (adapted from dashboard)
- rewrite controller to always authenticate requests
- add routes handling new flow
- add background SVG for login page

Here is a screenshot of Safari (top left), Edge (top right), Chrome (bottom left), Firefox (bottom right) showing the login page.

Screenshot 2023-02-13 at 23 36 06


line-o:
TIL that `git push -d` deletes a branch :/ Well, I recreated the PR #594

duncdrum (approved):
looks great thank @line-o

controller.xq
```diff
@@ -1,128 +1,159 @@
...
+(: public :)
+
 if ($exist:path eq '') then
     <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
-        <redirect url="{request:get-uri()}/"/>
+        <redirect url="{$local:uri}/"/>
     </dispatch>
 
-else if ($exist:path eq '/') then
+else if ($exist:path eq '/' and $user-allowed) then
```
line-o: `and $user-allowed` is not needed here (unauthenticated users will have an additional redirect)

controller.xq
```diff
@@ -1,128 +1,159 @@
...
-            else
-                (
-                    response:set-status-code(401),
-                    <status>fail</status>
-                )
-        } catch * {
-            response:set-status-code(401),
-            <status>{$err:description}</status>
-        }
+else if (
```
line-o: This route can be moved below handling of unauthorised requests

login.html
```diff
@@ -0,0 +1,134 @@
...
+        .submit-button:-moz-focusring { outline: 1px dotted ButtonText }
+
+    </style>
+</head>
+<body class="login-page">
+    <main class="login-box">
+        <h1 class="header">Sign in</h1>
+        <form action="login" method="post">
+            <input type="hidden" name="duration" value="P7D"/>
+            <input id="route" type="hidden" name="route"/>
```
line-o: this input is not needed here

joewiz:
I have checked out this branch and built and installed it into eXist 6.2.0. I can confirm that setting guest=no in eXide's configuration.xml file to cause any request to eXide to redirect to http://localhost:8080/exist/apps/eXide/login.html - with the page appearing as shown in the screenshots above. When guest=yes (as it is by default), this PR does not alter eXide's login UI at all. 

@line-o Is this intentional, or was the new login screen meant to replace the original login UI too?

One suggestion. The placeholder for the user input field `itsme` is unusual. I'd recommend changing this placeholder text to "Username..." Similarly, the placeholder for the password input ifeld `··········` makes it appear as if a password has *already* been entered, so I'd recommend changing this placeholder text to "Password..."

line-o:
@joewiz thank you for taking the time to review my PR.

> Does the login page replace the current login form?

Yes, when guest access is restricted it is unsafe to even show the eXide page. If you are logged in and this can see eXide the form will still work and allow to switch to a different user.

Nothing changes for instances where guest access is allowed.

> unconventional placeholders

I will change the placeholders as you suggested.

joewiz (approved):

Issue #651, opened 2023-12-19
[BUG] dbmanager spec is flaky


duncdrum:
**Describe the bug**

the spec is flaky because it doesn't reliably deal with the initial notice windows exide offers. So tests will fail because the elements that cypress is trying to interact with aren't visible or focused. 

**Expected behavior**
Robust tests with predictable behaviour

**Screenshots**
<img width="1512" alt="Screenshot 2023-12-19 at 22 27 13" src="https://github.com/eXist-db/eXide/assets/6205362/5b20a2d3-039c-40dd-ac3e-2ad2cf110980">

**Context (please always complete the following information):**
 - eXist-db Version: [6.2.0, 7.0.0-SNAPSHOT]
 - Java Version: [8, 17]
 -  Node Version: [18, 20]
-  App Version: [3.5.1]

joewiz:
Related: #652

joewiz:
Here's something strange that's perhaps related? @line-o's PR #711 passed CI, but the merge commit failed CI: https://github.com/eXist-db/eXide/actions/runs/10540075485. The failing tests involved dbmanager:

```
  19 passing (1m)
  2 failing

  1) DB Manager
       DB Manager operations
         cut operation
           should create collection to be copied:
     AssertionError: Timed out retrying after 4000ms: Expected to find content: 'toBeCopiedAéB' within the selector: 'div[role=gridcell][col-id=name]' but never did.
      at Context.eval (http://localhost:8080/__cypress/tests?p=cypress/integration/dbmanager.spec.js:354:75)

  2) DB Manager
       DB Manager operations
         cut operation
           should cut the collection:
     AssertionError: Timed out retrying after 4000ms: Expected to find content: 'toBeCopiedAéB' within the selector: 'div[role=gridcell][col-id=name]' but never did.
      at Context.eval (http://localhost:8080/__cypress/tests?p=cypress/integration/dbmanager.spec.js:381:14)
```

PR #711, opened 2024-08-20
[bugfix] outline view in sidebar

line-o:
The outline route returned statuscode 404 because it is in fact a post request and only get was allowed in the controller.

### DIFF
--- a/controller.xq
+++ b/controller.xq
@@ -240,7 +240,7 @@ else if ($local:method = 'get' and starts-with($exist:path, '/results/')) then
         </forward>
     </dispatch>
 
-else if ($local:method = 'get' and $exist:resource eq "outline") then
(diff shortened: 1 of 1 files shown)


line-o:
@dizzz here is the PR for the issue you reported yesterday.

dizzzz (approved):

joewiz (approved):
I can confirm the 404 in the current version and that the PR restores the feature. Thanks @line-o!

joewiz:
I'd like to issue a new release based on this but noted a CI failure after merging the PR - see https://github.com/eXist-db/eXide/issues/651#issuecomment-2308956313. I haven't had time to dig into this to see if it's a real problem or a problem related to this PR. I'd be interested in others' opinions.

line-o:
It is really just a flaky test that passes usually on the second run.

PR #764, opened 2026-03-06
Replace xqlint with REx parser and Prettier formatter

joewiz:
## Summary

- Replace eXide's bundled xqlint parser with a REx-generated parser built from the W3C XQuery 3.1 + Update 3.0 EBNF grammars
- Replace the disabled xqlint CodeFormatter with Prettier, supporting 8 languages
- Add 107 unit tests covering parser, visitor, utilities, and static analysis

See #765 for the broader modernization vision.

### REx parser (`src/parser/XQueryParser.js`, `src/parser/adapter.js`)
- REx-generated parser from W3C XQuery 3.1 + Update 3.0 grammars (to be replaced with 4.0 when the time is right)
- Adapter converts REx's TopDownTreeBuilder output to eXide's existing AST node shape
- All visitor infrastructure works unchanged

### Static analysis (`src/static-analysis.js`)
- Replaces xqlint's Translator for namespace, variable, and declaration checks
- Dead `Compiler` import removed from `src/quickfix.js`

### Prettier formatting (`src/prettier-format.js`, `tools/bundle-prettier.js`)
- esbuild bundles Prettier + plugins into a single browser IIFE (~2MB)
- Supports: XQuery, XML, HTML, CSS, Less, JavaScript, JSON, Markdown
- "Format code" moved to Edit menu (available for all supported modes)
- Plugins: prettier-plugin-xquery, @prettier/plugin-xml, built-in html/postcss/markdown/babel/estree

### Other changes
- `index.html.tmpl` switched to dev mode (individual source files)
- `build.xml` updated with new source files in concat list
- Format command registered in all mode helper constructors

## Remaining xqlint dependency

`XQueryLexer` in `src/ace-modes.js` — used as the Ace tokenizer. Will be removed in a future Ace → CodeMirror 6 migration.

## Test plan

- [x] `npm test` — 107 unit tests pass (parser, visitor, utils, static analysis)
- [x] `ant xar && xst package install` — builds and deploys successfully
- [x] XQuery formatting via Edit > Format code
- [x] XML formatting (with `xmlWhitespaceSensitivity: "ignore"`)
- [x] HTML, CSS, Less, JavaScript, JSON formatting
- [x] Error case: malformed code shows error, doesn't corrupt editor
- [x] Keyboard shortcut (Cmd-Shift-P / Ctrl-Shift-P) works across all modes


package.json
```diff
@@ -11,6 +11,8 @@
 		"build:dev": "node scripts/build.js prepare && npm run xqlint && node scripts/build.js --dev",
 		"clean": "node scripts/build.js clean",
 		"xqlint": "cd support/xqlint && node r.js -o build.js && cd ../..",
```
DrRataplan: Can disappear: no more xqllint

package.json
DrRataplan: build script(s) do not need to call xqlint anymore, but instead bundle-prettier

package.json
```diff
@@ -29,12 +31,15 @@
...
 	"homepage": "https://github.com/eXist-db/eXide#readme",
 	"devDependencies": {
+		"@prettier/plugin-xml": "^3.4.2",
 		"chalk": "^4.1.2",
 		"command-line-args": "^6.0.1",
 		"cypress": "^15.11.0",
 		"esbuild": "^0.27.3",
 		"glob-stream": "^7.0.0",
 		"micro-fs": "^1.0.3",
+		"prettier": "^3.8.1",
```
DrRataplan: not a devdep, but an actual one

package.json
DrRataplan: orr actually, call bundle-prettier as well

DrRataplan (commented):
I tried running this locally, but I am getting this error:

```
$ npm run build

> exide@3.5.5 build
> node scripts/build.js prepare && npm run xqlint && node scripts/build.js


> exide@3.5.5 xqlint
> cd support/xqlint && node r.js -o build.js && cd ../..

node:internal/modules/cjs/loader:1424
  throw err;
  ^

Error: Cannot find module '/home/martin/projects/exide/support/xqlint/r.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1421:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1059:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1064:22)
    at Module._load (node:internal/modules/cjs/loader:1227:37)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
    at node:internal/main/run_main_module:33:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v24.11.1

```

I have no clue where r.js should come from and why the CI accepts it.


This will start complaining about the exist db update expressions, because the parser only accepts XQuery 3.1 + XQuery update 3.0. I have an ebnf I use for the prettier-plugin-xquery stuff that handles XQuery 3.1 plus exist-db update expressions, but that does not do XQuery update... https://github.com/DrRataplan/xq-parser/blob/main/transformedgrammars/XQuery-31-full.ebnf

package.json
joewiz: [Response and reply here co-authored with Claude Code. -Joe]

Fixed in e07e6a2. Removed the `xqlint` script entirely and replaced `npm run xqlint` with `npm run bundle-prettier` in both `build` and `build:dev`. Also cleaned up the stale xqlint references in `scripts/build.js` (the `prepare()` copy step and `xqlint.min.js` in `clean()`).

package.json
joewiz: [Response and reply here co-authored with Claude Code. -Joe]

Good catch — fixed in eb84f3e. Moved `prettier`, `@prettier/plugin-xml`, and `prettier-plugin-xquery` to `dependencies` since they're bundled into `prettier-bundle.js` which ships in the .xar.

joewiz:
[Response and reply here co-authored with Claude Code. -Joe]

Thanks for testing this locally, Martin! The `r.js` error was because `npm run build` still called `npm run xqlint`, which tried to run RequireJS inside the `support/xqlint` submodule. Since xqlint has been replaced by the REx parser, that entire build step is dead code.

Fixed in three commits:
- e07e6a2: Remove the `xqlint` script and replace it with `bundle-prettier` in the build pipeline
- eb84f3e: Move prettier/plugin packages from devDependencies to dependencies
- 5e1420e: Delete the dead `scripts/xqlint/` directory (RequireJS build config)

Re: the eXist-db update expressions — the REx parser is built from `XQuery-Update-30.ebnf` which covers XQuery 3.1 + XQuery Update Facility 3.0. eXist-db's update syntax (`update insert/delete/replace/rename/value`) isn't in the W3C grammar, so those will flag as parse errors. Thanks for the pointer to your xq-parser grammar — that could be useful for a future pass to add eXist-db extension support.

joewiz:
@DrRataplan I'd be happy to fix the PR to avoid these separate fix commits, either before or after your review - whichever you'd prefer and helps most.

joewiz:
To facilitate cleaner views of these PRs, I've reproduced them in my fork, where I can chain the PRs (I didn't realize that chaining is only possible when the chained PRs are in the same remote). See this listing of the PRs, with links to each: joewiz#15. I think that perhaps it makes better sense to work in my fork until the present batch of work is finished and ready for final review. I'd welcome comments/reviews of the individual PRs there.

PR #774, opened 2026-03-07
Modernize eXide: CM6 editor, REx parser, UI redesign

joewiz:
## Summary

Comprehensive modernization of eXide. Replaces core infrastructure (editor, parser, formatter) and redesigns the UI.

Closes #765.

### Infrastructure
- **CodeMirror 6** replaces Ace editor — syntax highlighting, folding, bracket matching, search, lint gutter
- **REx-generated XQuery 3.1 parser** replaces xqlint — W3C grammar-based, with adapter layer preserving existing visitor infrastructure
- **Prettier integration** for multi-language code formatting (XQuery, XML, HTML, CSS, Less, Markdown)
- **jQuery/jQuery UI removed** — all AJAX, DOM manipulation, dialogs, and events converted to vanilla JS

### UI Redesign
- Modernized toolbar with action buttons, panel toggle SVG icons, dark mode cycle button
- CSS-mask logo tinted to match menu text color
- **Status bar** with live cursor position (clickable for Go to Line), APP segment, and TYPE popover selector
- Error status pill and slide-up detail panel with copy button (MutationObserver-based)
- **Outline view** with three display modes (nested/document, flat/document, flat/alpha), toolbar, and filter
- **Collections pane** with context menu, drag-drop, filtering, and file-type icons
- **Tab bar overflow** — scrollable tab strip with ‹/›/▾ controls and open-files dropdown menu
- **Binary file preview** overlay for images, PDFs, audio, and video
- Tab drag-and-drop with left/right insertion indicators
- Restyled dialogs: navy titlebar, consistent fonts/sizing, focus rings
- **DB Manager** file-type icons, toolbar refinements, browse filter fix
- **Layout preferences** — side panel, results panel, stacked view, results position
- Comprehensive dark mode with system theme detection

Screenshot 2026-03-09 at 00 15 06

Screenshot 2026-03-09 at 00 15 41

### Mode Helpers
- **CM6 syntax tree** for CSS, JavaScript, and Less outline extraction (replaces regex-based)
- **JSON mode helper** with key/value outline support
- **Markdown mode helper** with heading outline support
- XQuery outline entries now include row numbers for goto-line

### Cleanup
- Deleted dead files: `require.js`, `modernizr.custom.js`, `editor-shim.js`, `xqlint.min.js`, jQuery dir, black-tie CSS, pnotify CSS
- Removed dead CSS rules (`.ui-front`, `.ui-progressbar`)
- Removed dead debug code in `eXide.js` `startDebug()`
[...]

PR #778, opened 2026-03-15
Modernize eXide: CM6 editor, REx parser, REST API, LSP integration

joewiz:
## Summary

A comprehensive modernization of eXide, replacing legacy dependencies with modern alternatives and adding LSP-powered language intelligence. This consolidates work from multiple prior PRs (#774, #777) into a single reviewable PR with grouped commits.

### 1. Replace xqlint parser with REx-generated XQuery 3.1 parser
REx-generated parser from W3C EBNF grammar (XQuery 3.1 + Update 3.0 + Full Text 1.0). Adapter layer preserves the existing visitor infrastructure. Includes unit tests for parser, visitors, and static analysis.

### 2. Replace Ace editor with CodeMirror 6
Full migration from Ace to CM6 with language support, syntax highlighting, keybinding infrastructure, and editor-utils compatibility layer.

### 3. Add Prettier-based code formatting
Multi-language formatting via Prettier + plugins (XQuery, XML, HTML, CSS, Less, Markdown), bundled with esbuild. Replaces xqlint's CodeFormatter.

### 4. Remove jQuery, jQuery UI, and AG Grid dependencies
All AJAX calls use fetch, dialogs use native `<dialog>` via dialog-utils.js, DOM manipulation uses vanilla JS. AG Grid replaced with vanilla HTML tables.

### 5. Add OpenAPI/Roaster REST API and remove legacy XQuery modules
Replace 15+ legacy XQuery modules with a Roaster-based OpenAPI REST API. All backend operations (storage, query, packages, auth, sync, search, admin) handled through typed JSON endpoints.

### 6. Redesign UI
Toolbar, status bar with mode selector, native dialog system, tab overflow controls, filterable outline panel, dark mode with proper syntax highlighting (One Dark palette via CSS classHighlighter), system theme detection, keyboard shortcuts dialog, monitoring panel, and responsive layout.

### 7. Add CM6 autocomplete, semantic highlighting, and multi-language helpers
CM6 autocompletion with LSP-aware function completions (prefixed and unprefixed default namespace), semantic highlighting via decorations, hover tooltips with AST fallback, go-to-definition (same-file and cross-module), multi-cursor rename, and language helpers for JSON, Less, Markdown, JavaScript, CSS, and XML. Native HTML tag/attribute, CSS property/value, and JavaScript local variable completions via CM6 language packs.

### 8. Add Cypress integration tests and remaining frontend updates
E2E tests for autocomplete, diagnostics, outline, query execution, preferences, formatting, and more (172+ tests). Remaining frontend updates: layout, find/replace, menus, directory browser, drag-and-drop, build scripts, login page.

### 9. Add tag matching, exist-next compat, and CI fixes
- Matching tag highlight for XML/HTML (cursor on open tag highlights close tag)
- XML/HTML scope breadcrumb in status bar (XPath at cursor position)
- XML Rename Element with multi-cursor (both open + close tags)
- XPath signatures with positional predicates in outline and QuickPicker
- Navigation flash highlight on all jump-to actions
- Selection match highlighting, Cmd+D multi-select, diagnostics panel
- Autocomplete-on-type preference, dynamic namespace registry from server
- Local hover fallback, unprefixed default namespace completions
- Deferred splash screen, dialog focus on primary button
- exist-next compatibility: xs:dayTimeDuration casts, dynamic file module detection, package lookup fix
- CI: skip LSP-dependent tests when lsp:* module unavailable

### 10. Add Find All References
- Shift+F3 / Navigate > Find All References: QuickPicker listing all references to function/variable at cursor
- Server-side lsp:references() with AST visitor fallback
- Fix hover error handling for non-200 responses

### 11. WebSocket transport and monitoring
- WebSocket transport layer (`ws-transport.js`) for real-time LSP and monitoring push
- Monitor panel with WebSocket-based metrics, running/recent queries, kill query button
- LSP diagnostics push via WebSocket after compilation
- Auto-connect with exponential backoff reconnect

### 12. Server-side cursor query execution
Replaces the old XQueryServlet + session.xq flow with a cursor-based architecture using `lsp:eval`/`lsp:fetch`/`lsp:close` from the [exist-lsp](https://github.com/joewiz/exist-lsp) package.

**Before:**
```
User clicks Eval → HTTP POST to XQueryServlet → session.xq stores results
  → 10 GET requests for results/1..10 → pagination via more GETs
```

**After:**
```
User clicks Eval → POST /api/query → lsp:eval() stores result in Caffeine cache
  → GET /api/query/{id}/results?start=1&count=10&method=xml&indent=yes
  → lsp:fetch() serializes just that page with requested serialization params
  → changing serialization mode re-fetches current page (no re-execution)
```

Key properties:
- **Lazy serialization** — only the viewed page is serialized, with configurable [W3C serialization parameters](https://qt4cg.org/specifications/xslt-xquery-serialization-40/Overview.html)
- **No client-side buffering** — results stay on the server as live node references
- **Document links** — `documentURI` and `nodeId` available per result item
- **Dynamic serialization** — change output method, indent, or highlight-matches without re-executing
- **Pagination** — identical UX with next/previous/first/last

This follows the same pattern used by SQL clients (server-side cursors), BaseX GUI (in-process result references), and eXist's own Java admin client (XML-RPC result handles).

#### Cursor store cache policy

Results are held in a Caffeine cache with configurable eviction:

| Setting | Default | Description |
|---|---|---|
| `cursor.maximumSize` | 100 | Max concurrent cursors (LRU eviction when exceeded) |
| `cursor.expireAfterAccess` | 300000 (5 min) | Inactivity timeout in ms |
| `cursor.maximumWeight` | 0 (unlimited) | Max estimated total memory in bytes. When > 0, replaces count-based eviction. Weight per cursor = itemCount × 1KB. |

Configuration via `exist.xml` module parameters:
```xml
<parameter name="cursor.maximumSize" value="50"/>
<parameter name="cursor.expireAfterAccess" value="600000"/>
<parameter name="cursor.maximumWeight" value="536870912"/>
```

### 13. Query cancellation

**What it does:**
1. **Client-side**: `AbortController.abort()` drops the HTTP connection
2. **Server-side**: Kills running queries via `system:kill-running-xquery()` (admin only)

**Risks** — cancelling state-modifying queries can leave the database inconsistent:

| Expression type | Risk | Notes |
|---|---|---|
| XQUF (`insert`/`delete`/`replace`/`rename`) | Medium | PUL may be partially applied |
| Legacy update (`update value`, `update insert`) | High | Applied immediately, no batching |
| `xmldb:store()`/`xmldb:remove()` | High | Side effects immediate |
| `sm:chmod()`/`sm:chown()` | Medium | Permission changes immediate |
| `file:write()`/`file:delete()` | High | Cannot be rolled back |

**Mitigations:**
1. **Warning dialog** for queries containing update/store/delete/chmod/file expressions
2. **Admin-only kill** — guest users can only abort the HTTP connection (server query runs to completion)
3. **Preclaiming locks** ([eXist-db PR #6112](https://github.com/eXist-db/exist/pull/6112)) prevents deadlocks during partial update application
4. **Detection regex** covers XQUF, legacy update, `xmldb:*`, `sm:*`, `file:*` mutations
5. **Monitor sidebar** shows all running queries server-wide with per-query kill buttons

### 14. Results toolbar and menubar polish

- **Copy all results**: items are now separated by newlines when pasted
- **Number-of-results dropdown**: changing the page size now immediately re-fetches from page 1 (previously only updated the internal variable)
- **Active toggle buttons**: Indent and Highlight Matches buttons now show a solid blue background when active, clearly distinct from the hover state
- **Menubar toggle**: clicking an already-open menu header now closes it, consistent with standard menubar behaviour

## Temporary workarounds (remove after upstream merges)

- **LSP test skips**: Autocomplete and outline Cypress tests are skipped when lsp:* module is not available. Remove once eXist-db/exist#6130 is merged.
- **XQuery 4.0 Prettier guard**: Formatting rejects `xquery version "4.0"` code with a user-friendly message. Remove once prettier-plugin-xquery adds 4.0 support.
- **Dynamic file module import**: admin.xqm and sync.xqm use util:eval to dynamically load the file module. Can be simplified once eXist-db standardizes on the EXPath File module namespace.
- **Error line fallback**: compileError parses line/column from error message text when lsp:diagnostics() returns line 0. Remove once lsp:diagnostics() line fix is deployed.
- **function-lookup fallback**: query.xqm uses function-lookup for 4-arity lsp:fetch when the new overload isn't available. Remove once exist-lsp with serialization parameter support is deployed.

## Known limitations

- lsp:hover() does not return info for all registered module functions (e.g., sm:is-dba). Filed as server-side issue.
- XQuery 4.0 parser is lazy-loaded as a separate 664KB bundle to avoid doubling the main bundle size.
- Query evaluation is outside the [LSP spec's scope](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/) — `lsp:eval`/`lsp:fetch`/`lsp:close` are eXist-specific IDE support functions in the exist-lsp package. Reviewers are invited to weigh in on whether this scope is appropriate or if a separate package would be preferred.

## Requirements

- eXist-db 7.0+ with [exist-lsp](https://github.com/joewiz/exist-lsp) package installed
- Roaster 1.8.0+

## Test plan

- [x] Build and install XAR (`npm run build && xst package install local build/eXide-3.5.5.xar --force`)
- [x] Unit tests pass (`npm test` — 178 tests)
- [x] Cypress E2E tests pass (213 tests: 201 pass, 11 pending, 1 flaky monitor kill)
- [x] Full smoke test — 213 Cypress tests cover the vast majority; remaining visual/manual items tracked for future sessions
- [x] All LSP endpoints working — `autocomplete_spec` (11), `diagnostics_panel_spec` (5), `error_status_spec` (8), `native_autocomplete_spec` (7)
- [x] Cmd+Click go-to-definition with centered scroll and flash — `goto_references_spec` (F3 jump + clickable class)
- [x] Find All References (Shift+F3) with QuickPicker — `goto_references_spec` (QuickPicker opens + Escape closes)
- [x] Auto-triggered completions with prefix/unprefixed namespace support — `autocomplete_spec` covers prefixed, unprefixed, `fn:`, `repo:`, variable completions
- [x] Dark mode with One Dark syntax highlighting — `layout_spec` (dark mode layout)
- [x] Prettier formatting for XQuery 3.1, XML, HTML, CSS, JS, Markdown — `prettier_format_spec` (4 tests)
- [x] Outline panel with nested tree, XPath signatures, and filter — `outline_navigation_spec` (12 tests)
- [x] Tag matching, scope breadcrumb, selection match highlighting — `tag_scope_selection_spec` (CSS defined, XML/XQuery scope, selection match)
- [x] exist-next Docker compatibility verified — all development and testing done against exist-next Docker (eXist 7.0.0-SNAPSHOT)
- [x] Query execution with cursor pagination — `query_execution_spec` "paginates results with next/previous buttons"
- [x] Large result set (`1 to 10000`) — `query_execution_spec` "handles large result sets" (10 DOM nodes for 10K items)
- [x] Change serialization mode — `query_execution_spec` "re-fetches page when serialization mode changes without re-executing"
- [x] Toggle indent/highlight-matches — `query_execution_spec` "re-fetches page when indent toggle changes"
- [x] Cancel query with warning dialog — `query_execution_spec` cancel button tests (existence, visibility)
- [x] Monitor sidebar kill button — `monitor_spec` "shows running query and allows killing it" (flaky timing)


.github/workflows/ci.yml
```diff
@@ -40,14 +40,28 @@ jobs:
       - name: Build frontend Using node.js ${{ matrix.node-version }} and build EXPath Package
         run: npm run build
 
+      - name: Download Roaster dependency
```
duncdrum, line 42: I would recommend using gh cli, and download both latest release and the min version as per expath-pkg.xml in a matrix

cypress/e2e/lsp/error_status_spec.cy.js
```diff
@@ -0,0 +1,134 @@
...
+    cy.wait(50)
+  }
+
+  /**
+   * Wait for eXide to finish its initial compilation cycle.
+   * The editor compiles the default document on load, which writes to
+   * #error-status. We need to wait for that to settle before testing.
+   */
+  function waitForInitialLoad() {
+    cy.get('.path', { timeout: 10000 }).should('contain', 'untitled-1')
```
duncdrum, line 22: random timeouts are usually an anti pattern, is there something we can put on a named route, so we can wait on a request and its response?

cypress/e2e/lsp/error_status_spec.cy.js
```diff
@@ -0,0 +1,134 @@
...
+    // Wait for the initial compilation cycle to finish
+    cy.wait('@compile', { timeout: 10000 })
+    // Clear any existing error state before each test
+    cy.window().then((win) => {
+      const el = win.document.getElementById('error-status')
+      el.textContent = ''
+    })
+    cy.wait(50)
```
duncdrum, line 30: s.a.

cypress/e2e/autocomplete_spec.cy.js
```diff
@@ -0,0 +1,216 @@
...
+    }).then((resp) => {
+      lspAvailable = resp.status === 200 && Array.isArray(resp.body) && resp.body.length > 0
+    })
+  })
+
+  beforeEach(() => {
+    cy.loginXHR('admin', '')
+    cy.visit('/eXide/index.html')
+    cy.reload(true)
+    cy.get('.path', { timeout: 10000 }).should('contain', 'untitled-1')
```
duncdrum, line 22: s.a.

cypress/e2e/autocomplete_spec.cy.js
```diff
@@ -0,0 +1,216 @@
...
+    // Wait longer for login to propagate
+    cy.get('#user', { timeout: 15000 }).should('not.have.text', 'Login')
```
duncdrum, line 24: s.a.

cypress/e2e/autocomplete_spec.cy.js
```diff
@@ -0,0 +1,216 @@
...
+    if (!lspAvailable) this.skip()
+    setContentAndComplete('util:wai')
+
+    // Wait for async function completions to load
+    cy.get('.cm-tooltip-autocomplete', { timeout: 5000 })
+      .should('be.visible')
+      .invoke('text')
+      .should('contain', 'util:wait')
+
+    cy.wait(300)
```
duncdrum: s.a. I m not going to flag them all from here on out.

cypress/e2e/file_save_spec.cy.js
```diff
@@ -0,0 +1,142 @@
+describe('File save', () => {
+  // Use /db which always exists; clean up test files after
+  var testCollection = '/db'
+  var testFile = 'cypress-test-' + Date.now() + '.xq'
+
+  before(() => {
+    // Clean up stale test files from previous interrupted runs
+    cy.loginXHR('admin', '')
+    cy.request({
```
duncdrum: this is the second time we are cleaning up, which suggests its worth putting this into a custom command, so other specs can reuse it.

cypress/e2e/monitor_spec.cy.js
```diff
@@ -0,0 +1,96 @@
+describe('Monitor panel', () => {
+  beforeEach(() => {
+    cy.loginXHR('admin', '')
+    cy.visit('/eXide/index.html')
+    cy.reload(true)
```
duncdrum, line 5: this might no be necessary anymore, assuming the popup on first load situation is now under control

modules/api.json
duncdrum, line 1: IT would be nice to validate the schema on CI, to make future changes more manageable, similar to what we do in Jinks.

resources/schema/collection.xconf.xsd
duncdrum, line 1: we should not remove the license header here

duncdrum, line 1: woudn't if be nicer to dynamically load the schema from the existing instance?

resources/scripts/ui-init.js
duncdrum, line 1: most of these `var` can be `const` or `let`

scripts/build.js
duncdrum, line 1: just out of curiosity, and since going through all this trouble, do we still need `build.js`?  Why not use our own gulp?

src/deployment.js
duncdrum, line 1: is this covered in our tests?

src/dnd.js
duncdrum, line 1: lol for a moment I was wondering why we had a dungeons and dragons file

tools/REx.class
duncdrum, line 1: ?

tools/REx.java
duncdrum, line 1: Why do we store and commit binaries and java here?

expath-pkg.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://expath.org/ns/pkg" name="http://exist-db.org/apps/eXide" abbrev="eXide" version="3.5.5" spec="1.0">
```
duncdrum: `3.5.5` this should likely by `3.6.0` or `4.0.0`

package.json
```diff
@@ -7,11 +7,19 @@
...
 	},
 	"scripts": {
-		"build": "node scripts/build.js prepare && npm run xqlint && node scripts/build.js",
-		"build:dev": "node scripts/build.js prepare && npm run xqlint && node scripts/build.js --dev",
+		"build": "npm run bundle-cm6 && npm run bundle-prettier && node scripts/build.js prepare && node scripts/build.js",
+		"build:dev": "npm run bundle-cm6 && npm run bundle-prettier && node scripts/build.js prepare && node scripts/build.js --dev",
 		"clean": "node scripts/build.js clean",
-		"xqlint": "cd support/xqlint && node r.js -o build.js && cd ../..",
-		"cypress": "cypress run"
+		"test": "node --test test/parser-test.js test/visitor-test.js test/utils-test.js test/static-analysis-test.js test/snippet-test.js test/parser-registry-test.js test/xq40-syntax-test.js",
```
duncdrum: I think `node --test` should suffice here

duncdrum (commented):
I m starting to be a bit concerned about a mix of Java, Rust, Js,… it adds quite a bit of complexity. 

That being said love tauri. 

Should be build on ci though.

src-tauri/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "app"
+version = "0.1.0"
+description = "A Tauri App"
+authors = ["you"]
+license = ""
+repository = ""
+edition = "2021"
+rust-version = "1.77.2"
```
duncdrum, line 9: this needs to be on CI and on the readme

.github/workflows/ci.yml
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in commits ea7d771–1ddca80: replaced `curl` with `gh release download` (using the built-in `GITHUB_TOKEN`) for the Roaster dependency. The existing `release`/`latest` Docker matrix covers the most important cases — testing against the specific `semver-min` from `expath-pkg.xml` (6.1.0) would require an old Docker image that predates the Roaster/lsp:* dependencies this PR needs. Happy to add that as a separate matrix entry once we settle on the minimum version the full feature set requires.

cypress/e2e/lsp/error_status_spec.cy.js
joewiz: [This response was co-authored with Claude Code. -Joe]

Line 22 is already `cy.wait('@compile', { timeout: 10000 })` — a named intercept alias set up in `beforeEach` via `cy.intercept('POST', '**/api/query/compile').as('compile')`. The remaining `cy.wait(50)` at line 30 is waiting for a browser MutationObserver microtask to fire after writing to `#error-status`, which can't be replaced with a network route. It's a minimal delay for a synchronous DOM write, not a timing guess.

cypress/e2e/lsp/error_status_spec.cy.js
joewiz: [This response was co-authored with Claude Code. -Joe]

Same as above — this is the MutationObserver microtask wait (50ms), not a network timing guess.

cypress/e2e/autocomplete_spec.cy.js
joewiz: [This response was co-authored with Claude Code. -Joe]

These lines are `cy.visit()` and `cy.get('.path').should('contain', 'untitled-1')` — no `cy.wait()` here. The numeric waits that were the real issue (lines 95/100/135/163/168/195/200) have been replaced with `cy.intercept('POST', '**/api/query/completions').as('completions')` + `cy.wait('@completions')` in commit 24ef119.

cypress/e2e/autocomplete_spec.cy.js
joewiz: [This response was co-authored with Claude Code. -Joe]

These lines are `cy.visit()` and `cy.get('.path').should('contain', 'untitled-1')` — no `cy.wait()` here. The numeric waits that were the real issue (lines 95/100/135/163/168/195/200) have been replaced with `cy.intercept('POST', '**/api/query/completions').as('completions')` + `cy.wait('@completions')` in commit 24ef119.

cypress/e2e/autocomplete_spec.cy.js
joewiz: [This response was co-authored with Claude Code. -Joe]

Fixed in commit 24ef119: all `cy.wait(300/200/500)` in the autocomplete tests are replaced with `cy.wait('@completions')` / `cy.wait('@completions2')` on named intercept aliases. `acceptCompletion` dispatches a synchronous CM6 transaction, so the trailing `cy.wait(200)` was also removable.

cypress/e2e/file_save_spec.cy.js
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in commit 42e6a71: extracted to `cy.cleanupTestFiles()` in `cypress/support/commands.js`. Both `before()` and `after()` now call it, and other specs that create files in `/db` can reuse it.

cypress/e2e/monitor_spec.cy.js
joewiz: [This response was co-authored with Claude Code. -Joe]

The admin login is still needed here: the `app.login.isAdmin` test, the `api/admin/status` test, and the kill-query test all require an authenticated admin session. Happy to revisit if you saw a specific scenario where it caused interference.

modules/api.json
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in commit 1ddca80: added `npx --yes @redocly/cli lint modules/api.json` as a CI step before the build. Using `npx --yes` avoids adding it as a dev dependency.

resources/schema/collection.xconf.xsd
joewiz: [This response was co-authored with Claude Code. -Joe]

Restored in commit da20fc6. The header was accidentally dropped when syncing the schema from eXist-db/exist.

joewiz: [This response was co-authored with Claude Code. -Joe]

Interesting idea — the editor could fetch the schema from the running instance rather than shipping a static copy. The tradeoff is it would require the instance to be reachable (no offline editing of collection configs) and adds a startup request. Worth tracking as a follow-up.

resources/scripts/ui-init.js
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in commit 17c172b.

scripts/build.js
joewiz: [This response was co-authored with Claude Code. -Joe]

Yes, still needed — it handles three things: (1) esbuild bundling via the dedicated tool scripts, (2) template substitution (`{{version}}` in `expath-pkg.xml.tmpl` and `index.html.tmpl`), and (3) XAR packaging via `micro-fs`. It could be migrated to gulp in a follow-up, but the current footprint and build time are already low.

src/deployment.js
joewiz: [This response was co-authored with Claude Code. -Joe]

Not yet — the deployment dialog involves a multi-step UI (selecting server, entering credentials, triggering package install) that's harder to automate without a test server with the public repo accessible. Tracked in the future test gaps list.

src/dnd.js
joewiz: [This response was co-authored with Claude Code. -Joe]

`dnd.js` is for Drag and Drop, not Dungeons and Dragons — though XQuery dungeon crawling is admittedly an underexplored genre.

tools/REx.class
joewiz: [This response was co-authored with Claude Code. -Joe]

Removed in commit ea7d771. The `.class` file is a compiled Java binary for the REx parser generator — it shouldn't have been in version control. `generate-parser.js` now compiles it from `REx.java` on demand via `javac` if the class file is absent.

tools/REx.java
joewiz: [This response was co-authored with Claude Code. -Joe]

`REx.java` is the source for the REx parser generator (by Gunther Rademacher, bottlecaps.de), included for transparency and reproducible builds — the site hosts the tool but isn't always reachable. The compiled `REx.class` has been removed from git (commit ea7d771); `npm run generate-parser` now compiles it on first use via `javac`. The `.java` source stays so contributors can inspect and rebuild without depending on an external download.

expath-pkg.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

Bumped to `3.6.0` in commit a23c9cf. This is a significant feature release but preserves installation compatibility with the existing `expath-pkg` / `repo:install` flow, so a minor version bump feels right.

package.json
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in commit 1ddca80: simplified to `node --test`. Node 20+ auto-discovers `*-test.js` files in `test/`.

src-tauri/Cargo.toml
joewiz: [This response was co-authored with Claude Code. -Joe]

Agreed. Tauri CI requires platform-specific runners (macOS for `.dmg`, Windows for `.msi`, Linux for `.AppImage`) and signing credentials, which is a meaningful addition to the workflow. Tracking that as a follow-up PR focused on the desktop app. For this PR the Tauri scaffolding is present and buildable locally with `npm run build:desktop`; the README update and CI matrix will come with the desktop-app PR.

line-o:
@joewiz is it possible to split this PR up into 

1. add features and modernize application XAR 
2. eXide desktop

I do like the direction you are heading in and I would like to get this into the next release as a bundled application.
But for this we need to be able to grasp the changes.
A desktop application will increase the complexity and maintenance effort at this point that will put this goal at risk.
The desktop application can even be a separate project that has this repo as a  dependency.

joewiz:
@line-o Thanks for considering this PR for the next release! Sure, I can update the PR to set aside the eXide Desktop feature.

joewiz:
Superseded by #794, which contains only the web-app modernization (no eXide Desktop). The desktop scope will follow in a separate PR. Closing this PR but leaving the `pr/modernize-eXide` branch in place since it still holds the desktop commits needed for the follow-on PR.

Issue #25, opened 2013-07-05
Outline view doesn't differenriate argument differences with similarly-named functions


vadi2:
If I have both:

```
declare function a( $something as xs:string )
declare function a( )
```

Then both variations of a appear in the outline view, but clicking on the second one will always bring you to the first - it appears to only be matching by function name and not the arguments as well.

PR #794, opened 2026-05-11
Modernize eXide: CM6 editor, REx parser, Roaster API, LSP integration

joewiz:
## Summary

Modernizes the eXide web application, replacing core dependencies and adding language-service-powered intelligence, real-time monitoring, and a REST API. This PR contains **only the web-app modernization** — the eXide Desktop (Tauri) work that was part of #778 is deferred to a follow-on PR per @line-o's request.

Reviewers familiar with eXide pre-modernization and the goals discussed in #778 should find this PR much easier to follow: 19 topical commits instead of 56, no desktop scope, end-state dependencies ([`existdb-openapi`](https://github.com/eXist-db/existdb-openapi), `roaster`) assumed final.

## Highlights

- **REx parser** generated from W3C XQuery 3.1 EBNF replaces xqlint; visitor infrastructure unchanged.
- **CodeMirror 6** replaces Ace, with new CM6 autocomplete, semantic highlighting, and language helpers for XQuery, XML, HTML, CSS, Less, JS, JSON, Markdown.
- **Prettier** integration (XQuery, XML, HTML, CSS, Less, JS, JSON, Markdown).
- **jQuery, jQuery UI, AG Grid removed** — vanilla JS + native `<dialog>` via `dialog-utils.js`.
- **OpenAPI/Roaster REST API** replaces legacy XQuery modules.
- **Redesigned UI**: toolbar, status bar, dialogs, tabs, dark mode, outline.
- **Cypress integration tests** (autocomplete, monitor, dialogs, file save, go-to-definition, references, tag matching, scope, selection match, query execution, pagination, large result, serialization, WebSocket).
- **WebSocket transport** for real-time diagnostics + monitoring push (running query / recent queries, kill query). Diagnostics payloads follow LSP shape (e.g. `textDocument/publishDiagnostics`); see also [`existdb-openapi`](https://github.com/eXist-db/existdb-openapi) for the underlying language-service endpoints.
- **Server-side cursor pagination** via `cursor:eval` / `cursor:fetch`, with AbortController cancel, update-query warning, and adaptive serialization.

## Dependencies

- **eXist-db `develop`** (eXist 7.0+) with eXist-db/exist#6145 (Jetty 12 / WebSocket support, **merged**). CI runs against `existdb/existdb:latest`; the eXist-db release-version matrix entry is dropped because eXide 4.0 requires the 7.0 `lang:*` / `cursor:*` modules from existdb-openapi, which are only available against develop.
- [`existdb-openapi`](https://github.com/eXist-db/existdb-openapi) (formerly `exist-api`) — provides the `lang:*` (language services: diagnostics, completions, hover, definition, references, symbols) and `cursor:*` (server-side cursor: eval, fetch, close) functions consumed by eXide. Replaces the standalone `exist-lsp` package. CI downloads the release XAR like roaster.
- `roaster` ≥ 1.8.0 — Roaster REST API framework.

## Test plan

- [x] `npm test` — Node unit tests (179 tests) ✅ verified in [latest CI run](https://github.com/eXist-db/eXide/actions/runs/25984899778) (`# tests 179 # pass 179 # fail 0`)
- [x] `npm run build` produces `build/eXide-4.0.0.xar` ✅ verified in [latest CI run](https://github.com/eXist-db/eXide/actions/runs/25984899778) (`Creating xar eXide-4.0.0.xar`)
- [x] Install into eXist + `existdb-openapi` + `roaster`; eXide loads at `/exist/apps/eXide/`
- [x] `npx cypress run` — full E2E suite
- [x] Smoke test in browser: new file, save, run query, paginate large result, cancel, WebSocket monitoring, language-service diagnostics, go-to-definition, find references, format, dark mode, tag matching, outline

> **Note**: This PR's call sites originally used the `lsp:` XQuery namespace and `/api/lsp/*` URL prefix (when the upstream package was `exist-api`). Both were renamed in [existdb-openapi#18](https://github.com/eXist-db/existdb-openapi/pull/18): `lsp:` → `lang:` for language-service functions, plus a new `cursor:` namespace for `eval`/`fetch`/`close`. The CSS class `cm-lsp-hover` and helper file `src/lsp-hover.js` were renamed to `cm-lang-hover` / `src/lang-hover.js` to match.

Supersedes #778.


duncdrum:
needs a rebase

line-o:
I spent about 15 minutes working with this new version of eXide.
First of all: Thank you for putting so much effort into this. I think some very important choices were made which will make our lives easier going forward.
I also have to report following issues that let me hesitate to just give this an approval:

- the outline panel does not seem to do anything
- the function definition pop-up seems to only work in chrome
  - the pop-up also has full page-width and is therefore hard to read (tested with `matches`, see screenshot)
- in Safari I cannot type in the editor (that's a blocker)
  - update: I succeeded in getting this state in chrome as well
- I would prefer the editor window to have a white background in light-mode / this would make it easier for me to focus on the most important info on the screen

Screenshot 2026-05-29 at 09 25 10

line-o:
I got the editor in a state where it showed me a mysterious error about a ? somewhere invisible in my source and the only way to get rid of it was by reloading the website.

Screenshot 2026-05-29 at 09 38 15

line-o:
In Safari I see constant WebSocket traffic on `/eval` and on Chrome only POST requests to `/compile`

Screenshot 2026-05-29 at 09 59 35

line-o (changes requested):
There UI and usability issues need to be addressed before we can merge this PR.

duncdrum (commented):
Overall this is a huge improvement. There are a few outstanding issues though:

1. Logger noise
   - Compare to the severity of system events, the cursor store should be debug or dropped. There are multiple lines just from opening eXide via dashboard without any interaction
```
29 May 2026 07:30:24,323 [qtp527629265-51] INFO  (Deployment.java [installAndDeploy]:258) - Deploying package http://exist-db.org/apps/eXide 
2026-05-29 09:34:59

29 May 2026 07:34:59,990 [qtp527629265-49] INFO  (CursorModule.java [<init>]:61) - Cursor store: maximumSize=100, expireAfterAccess=300000ms, maximumWeight=unlimited 
```

   - after running the tests 
   
   ```
   29 May 2026 07:51:46,310 [qtp1531812826-61] WARN  (JakartaWebSocketFrameHandler.java [onError]:314) - Unhandled Error: org.exist.xquery.functions.websocket.WebSocketEndpoint@14c4507c 
2026-05-29 09:51:46

java.nio.channels.ClosedChannelException
2026-05-29 09:51:46

	at org.eclipse.jetty.websocket.core.internal.WebSocketSessionState.onEof(WebSocketSessionState.java:173)
2026-05-29 09:51:46

	at org.eclipse.jetty.websocket.core.WebSocketCoreSession.onEof(WebSocketCoreSession.java:235)
2026-05-29 09:51:46

	at org.eclipse.jetty.websocket.core.WebSocketConnection.fillAndParse(WebSocketConnection.java:543)
2026-05-29 09:51:46

	at org.eclipse.jetty.websocket.core.WebSocketConnection.onFillable(WebSocketConnection.java:380)
2026-05-29 09:51:46

	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:323)
2026-05-29 09:51:46

	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
2026-05-29 09:51:46

	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
2026-05-29 09:51:46
   ...
   ```

  I've seen similar errors when working on monex, this might be an openapi issue

2. CI config (decreasing severity)
    - Node version on CI should be `24` or `LTS/*` moving target, `22` is EOL
    - roaster step is no longer necessary since it ships with exist, downloading and installing on CI only makes sense if we want to have forward looking tests here against upcoming roaster releases,
    - same for existdb-openapi
    - few outdated deps `npm outdated` (nice to have)
    - add groupings to `dependabot.yml` for `@codemirror`, `prettier`,  `*` for GitHub actions, and cleanup outdated `ignore`  (nice to have)
3. expath packaging 
   - update dep version in `expath-pkg.xml.tmpl` processor `6` is intentional?
   - update changelog in `repo.xml`
   - why is there a `mods` namespace declaration in `collection.xconf` why is there an index definition at all?
4. This might be just on my system, if it's not we should drop `support/ace` from the index maybe even `support/` or what does it do?
5.  Test related (ran on Firefox)
    - `shows running query and allows killing it` on `monitor_spec` succeeds but ends with a `Server Error` message that I m not sure should happen
    - the TODO in `cypress/e2e/outline_modes_spec.cy.js` can be addressed, all cases are skipped, despite the feature being there. 
    - same with `cypress/e2e/outline_spec.cy.js`
    - during test execution cypress creates multiple copies of `/db/cypress-test-[some digits].xq` files in the db, it should clean up after itself.
6. Minor nitpicks (probably me holding it wrong)
   - I use the docs testsuite for playing with the `XQuery > Run as test` functionality, and couldn't get eXide to display the full error. The table view is nice, but when stuff breaks I want / need the full error output or junit report.

duncdrum:
@joewiz jyst checking i adjuste `evalWebsocketEndpoint` in the exist hotfix above. That is not the eval endpoint exude is connecting to?

joewiz:

Thanks both for the detailed review. Worked through everything; here's the per-item accounting. CI is green on the current tip. Real-Safari + Firefox validation completed manually today via the macOS Claude desktop app in Cowork (Safari, Firefox) and Cypress (Chrome + Firefox engines, full suite).

## @line-o's items

| Item | Status | Notes |
|---|---|---|
| **Outline panel does not seem to do anything** (Firefox) | ✅ **Cannot reproduce** | Cypress against Firefox 151.0.2: 9/9 outline tests pass; full 242-test suite green. Real-Firefox verified manually today: opening `/db/apps/eXide/modules/config.xqm` populates the outline tab with `config:repo-descriptor`, `config:expath-descriptor`, `config:get-configuration`, etc., and clicking an item navigates the editor to the function. If you can share the specific document that didn't work, happy to dig in. |
| **Function-definition popup full page-width, hard to read** | ✅ **Fixed** (`ae06014`) | The inner `.cm-funcdoc-tooltip` had `max-width: 700px`, but CM6's outer `.cm-tooltip` wrapper had no width constraint — it was free to stretch arbitrarily wide. Added a `.cm-tooltip:has(> .cm-funcdoc-tooltip)` selector that caps the outer wrapper at 550px (down from 700 per your readability feedback) and added `width: max-content` so the popup shrinks to fit narrower content. `:has()` is supported in Chrome 105+, Firefox 121+, Safari 15.4+. |
| **In Safari I cannot type** | ✅ **Fixed** (`8ad2913`) | Was a downstream symptom of the WebSocket storm — see next item. After the fix, typing into the editor in real Safari works (validated today: `(: hello from safari :)` typed successfully into the editor; cursor advanced to Col 45). |
| **Constant WebSocket traffic on /eval in Safari** | ✅ **Fixed** (`8ad2913`) | Root cause: `src/ws-eval.js` was auto-connecting at app init to `/exist/ws/eval`, an endpoint that only exists on an unreleased `feature/websocket-eval` branch of eXist-db. The server accepted the WS upgrade and immediately closed; the client's `onopen` reset the reconnect backoff to 1s, `onclose` scheduled a 1s reconnect, infinite loop. Verified in Playwright WebKit before the fix: 50+ open/close cycles in 10 seconds. After the fix: a single persistent `ws://localhost/exist/ws` connection (the monitoring channel). Chrome/Firefox tolerated the storm silently; Safari's stricter resource handling made it user-visible as the typing freeze. Two changes: disabled the autoConnect call (no consumer used `wsEval.execute()` anyway), and added a max-5-attempt reconnect cap with first-message-gated backoff reset so any future re-enablement against a broken endpoint can't sustain the loop. Real-Safari verification today: Web Inspector → Network → WS shows exactly one persistent connection, no churn. |
| **Mysterious error about a ? somewhere invisible, only fixable by reload** | ✅ **Fixed** (`32744fc`) | The existing `×` button only hid the error panel — the red pill in the status bar stayed visible, with no way to clear it short of reloading. Added an explicit "Dismiss" button to the error panel that clears the error entirely (cascades through the existing `#error-status` MutationObserver). The `×` keeps its existing "hide panel but keep pill" semantics for users who want to glance at vs. resolve. Real-browser verified today: error appears → Dismiss button is present alongside Copy and × → click Dismiss → panel closes and pill is gone. (Note: the per-line gutter marker remains, which is correct — it's a separate annotation surface tied to the underlying validator output.) |
| **Editor background should be white in light mode** | ✅ **Fixed** (`c890309`) | Added a CM6 `EditorView.theme(..., { dark: false })` extension that sets `&` (root editor) and `.cm-gutters` to `#ffffff` in the light-theme branch of `setTheme()`. Done as a CM6 theme extension rather than a CSS override on `.cm-editor` so it's colocated with the existing theme-switching logic. |

## @duncdrum's items

### 1. Logger noise

| Item | Status | Notes |
|---|---|---|
| **CursorModule INFO log on every request** | ✅ Released as **existdb-openapi v0.9.2** (your [#25](https://github.com/eXist-db/existdb-openapi/pull/25)) | Demoted to DEBUG. v0.9.2 is on public-repo; eXide CI workflow updated to v0.9.3 in `fec1677` (v0.9.3 also includes a changelog backfill — explanation below). |
| **JakartaWebSocketFrameHandler ClosedChannelException WARN** | ⏳ **Tracked separately** | Server-side issue in existdb-openapi / eXist core, not eXide. Worth a follow-up issue against existdb-openapi but out of scope for eXide#794. |

### 2. CI config

| Item | Status | Notes |
|---|---|---|
| **Node 22 EOL → 24** | ✅ Done (`fec1677`) | |
| **Drop roaster download** (ships with eXist) | ⚠️ **Held** | Verified `existdb/existdb:latest` doesn't yet ship roaster in `/exist/autodeploy/`; CI still downloads the v1.12.0 XAR. Easy follow-up cleanup once eXist bundles it. |
| **Drop existdb-openapi download** | ⚠️ **Held** | Same situation — not bundled in the eXist image yet. |
| **Outdated deps** | ✅ Done (`4e10313`) | `npm update` applied. Bumped @codemirror/* (8 packages), prettier 3.8.1→3.8.3, prettier-plugin-xquery 1.5.0→1.6.1, ws 8.20.0→8.21.0. Deferred chalk 4→5 and glob-stream 7→8 (major bumps with breaking changes — separate consideration). |
| **dependabot.yml: groupings + drop outdated ignores** | ✅ Done (`46250c3`) | Grouped `@codemirror/*` + `@lezer/*`, `prettier*`, and `*` for github-actions so related minor/patch bumps land as one PR per week. Removed the stale `cypress 6.x/7.x` ignore list (project is on 15.x). |

### 3. EXPath packaging

| Item | Status | Notes |
|---|---|---|
| **processor 6 is intentional?** | ✅ Bumped to `7.0.0-SNAPSHOT` (`5d96094`) | The 6.1.0 was inherited from eXide 3.x. eXide 4.0 actually requires eXist 7.0+ (Jetty 12 WebSocket, lang/cursor modules, Roaster ≥1.8). Used `7.0.0-SNAPSHOT` rather than bare `7.0.0` because EXPath / SemVer treats pre-release labels as ordering before the un-suffixed version, so `7.0.0` strict would refuse to install on current `7.0.0-SNAPSHOT` development builds. Tighten to bare `7.0.0` once 7.0.0 final ships. |
| **Update changelog in repo.xml** | ✅ Added 4.0.0 entry (`91f7ea7`) | Comprehensive entry summarizing the modernization landed at top of `<changelog>`. Also caught and back-filled the existdb-openapi changelog separately as **v0.9.3** since v0.9.0/v0.9.1/v0.9.2 had no `<change>` entries (caused exist-db.org/public-repo to render an empty Change log panel for that package). |
| **collection.xconf — mods namespace + index def, why?** | ✅ **Deleted both** (`601d55a`) | Per Joe: file has been there ~15 years and performs no function. Removed `collection.xconf` (the no-op fulltext-disable rule with the vestigial mods namespace), `pre-install.xq` (whose only purpose was installing collection.xconf), and the `<prepare>pre-install.xq</prepare>` reference in repo.xml. |

### 4. support/ace cleanup

| Item | Status | Notes |
|---|---|---|
| **support/ace lingering** | ✅ Already absent | No `support/` directory exists in the source tree; the Ace removal cleanup was complete. The `!support/**` exclusion in scripts/build.js is defensive, leaving no harm. |

### 5. Test-related

| Item | Status | Notes |
|---|---|---|
| **monitor_spec ends with "Server Error" toast** | ✅ **Fixed** (`1357054`) | When the monitor's kill button DELETEs the running query, the server immediately drops its in-flight POST /api/query response connection. The fetch on the eXide tab rejects with a generic `TypeError: Failed to fetch` (Chrome/Firefox) or `Load failed` (Safari) — NOT an `AbortError`. The catch in eXide.js treated it as an unexpected server bug. Detect the network-error shape and treat it as a cancellation (toast: "Query cancelled.") instead. |
| **outline_spec / outline_modes_spec all skipped** | ✅ **Re-enabled** (`afb502d`) | Removed the runtime `lspAvailable` probe + skip guards from both files; the LSP endpoint is now reliably present in CI because the workflow downloads existdb-openapi from the org release. 9 previously-skipped tests now actively pass. |
| **cypress-test-*.xq files not cleaned up** | ✅ **Fixed** (`efc9798`) | Added a session-level `after()` in `cypress/support/e2e.js` that calls `cy.cleanupTestFiles()` once after all specs finish. Catches files left behind by specs that failed mid-test, older runs predating per-spec cleanup, and any future spec that creates files and forgets a cleanup hook. |

### 6. Run-as-test display

| Item | Status | Notes |
|---|---|---|
| **Need full error output / JUnit report** | ✅ **Per-row expand** (`d17753c`) | Failed test rows now render the short `.message` as a `<summary>` inside a `<details>` element, with the full `.detail` (XQuery stack trace, etc.) in an expandable `<pre class="test-full-output">` below. Capped at 300px with overflow:auto for very long traces. CSS includes dark-mode counterparts. Deferred the full JUnit-XML export — happy to follow up in v4.1 if it's still wanted after using the per-row expand. |

## Other items resolved during this round

- **WebSocket pings** — answered in Slack: text "ping" frames are old cruft from the eXist-db WebSocket framework, not intentional client behavior. Native WS Ping/Pong is handled transparently. The right fix lives server-side; the client-side guard can be deleted once those frames stop. Duncan has the context.
- **autocompleteOnType** preference: default flipped to false to match pre-modernization eXide UX, with new tests verifying the toggle behavior + persistence (`7cd03dc`, `f0a0607`).
- **Goto-references test was Mac-only**: fixed to dispatch the correct modifier key per platform (`f0e26d9`) — Linux CI needed `Control`, macOS needs `Meta`.
- **Multiple smaller fixes** along the way: completions latency (server-side fallback was running for unprefixed prefixes), monitor false-positive running-query count (`system:query` → `system:xquery` + self-filter).

## What I haven't done

- **JUnit XML export for Run as test** — per-row expand should cover the practical use case; happy to add JUnit as v4.1 if needed.
- **Drop roaster/existdb-openapi from CI autodeploy** — held until those packages ship in the `existdb/existdb:latest` image.
- **Cross-platform Cypress CI matrix** (adding macos-latest alongside ubuntu-latest) — separate decision; would catch the kind of Mac-vs-Linux test divergence I hit earlier in this PR, but adds ~10× the runner cost.

## Verification

- **CI green** on latest commit (`32744fc`) — eXist-db CI Linux build
- **Local full Cypress suite green** against `existdb/existdb:latest` + existdb-openapi 0.9.3: 242 tests, 240 passing, 2 pending (Chrome AND Firefox)
- **Real-Safari + real-Firefox manual validation today**: typing works, no WebSocket storm (1 persistent connection only), outline panel works, error panel Dismiss button works as designed
- **Playwright WebKit** verified the engine-level Safari fixes before manual Safari validation

Ready for re-review when you have a moment 🙏

joewiz:

@duncdrum — yes, same endpoint. eXide's `src/ws-eval.js:117` builds the URL as `<scheme>://<host>/exist/ws/eval`, and your `EvalWebSocketEndpoint.java` is annotated `@ServerEndpoint(value = "/ws/eval")`.

Worth noting that on PR #794 I also commented out the `wsEval.autoConnect()` call (`8ad2913`) since nothing in eXide currently uses `wsEval.execute()` — so eXide 4.0.0 won't actually open the connection at all. The two fixes are complementary: yours makes the server-side endpoint robust for whatever client uses it (now or future), mine keeps eXide from opening it speculatively until there's a real consumer. Thanks for the hotfix.

duncdrum (approved):

line-o (changes requested):
Looks better now also in Safari although I got really weird display issue when radically resizing the browser window.

I can confirm that the websocket storm is fixed and I can type in the editor window.
The outline in Safari for a main module still does not seem to work reliably.

Another showstopper though: the simplest of queries `count(//p)` fails with an "Unknwon Error"

The same query works as expected on the same instance when run with `xst run "count(//p)"`

Image

joewiz:

@line-o — thanks for catching this. The "Unknown error" was hiding a much more useful message. Tracked the root cause to a field-name mismatch in eXide's error-response parser:

`existdb-openapi`'s `/api/query` endpoint returns errors as:

```json
{ "code": "err:XPST0081",
  "description": "It is a static error if a QName ...",
  "line": 1, "column": 3, "module": null, "value": null }
```

…where the human-readable message is in `description`. But eXide was looking for `err.error` (a legacy shape used by older eXist REST endpoints) and falling through to the literal string `"Unknown error"`. So the actual server-side error was thrown away.

Fixed in `156e994` — eXide now coalesces `description` → `error` → `message` → stringified payload, surfaces the error code in brackets, and falls back to `HTTP <status>: <body>` if the response body isn't even JSON. There's also a new regression test (`cypress/e2e/query_error_display_spec.cy.js`, 3 cases) that submits queries that any conformant XQuery engine will reject and asserts the real error text appears.

**Could you re-run `count(//p)` against the new build?** With this fix, instead of "Unknown error" you'll see the actual server response — likely an `err:XPDY0002` (no context defined) or similar. That'll tell us whether the underlying cause is something to fix in eXide, in existdb-openapi, or in eXist core.

For the other two items in your review:

- **Safari display issue on radical window resize** — happy to investigate if you can share roughly what happened (e.g., panels disappeared, editor scrollbar broke, etc.). Likely a CSS / layout thing, not the modernization core.
- **Outline in Safari for a main module not reliable** — testing here in real Safari today via library modules (`config.xqm`) showed the outline populating correctly. If you can share which main module showed the issue, I'll try the same.

All three items can be follow-ups against the eXide repo (don't have to block this PR) — but obviously the count(//p) message issue had to land here since the parser bug was real and confusing.

CI: green on the new tip. Local Cypress: 245 tests, 243 passing, 0 failing (the new spec added 3).

line-o:
The thing is that `count(//p)` should return 66 and not raise any error.

joewiz:

@line-o — you were right; thank you for pushing back. `count(//p)` should return the actual count, not 0, and not raise an error. I owe you an apology for the previous round: I shipped a fix for only the error-display half and asked you to re-verify, when I should have run `count(//p)` myself, compared it to the REST baseline, and caught the deeper defect before responding. That's on me, not you.

The underlying bug was in **existdb-openapi**, not eXide: `cursor:eval` never called `setStaticallyKnownDocuments` or `setBaseURI`, so unprefixed path expressions (`//p`) had no documents to walk. And switching to the right setup also required dropping `copyContext()` in favor of a fresh `XQueryContext` — `copyContext` caches the parent's resolved `staticDocuments`, which the path-expression optimizer then captured at compile time. (More detail in the PR.)

The PR with the fix: https://github.com/eXist-db/existdb-openapi/pull/26

Verified end-to-end through the eXide → `/api/query` → `cursor:eval` path on a local 7.0.0-SNAPSHOT container — `count(//p)` against `/db/apps/eXide` now returns 36, matching `/exist/rest/db?_query=count(//p)` exactly. Three regression tests added (`src/test/cypress/e2e/query_scope.cy.js`) modeled on `RESTServiceTest.queryGet()` so cursor:eval's coverage mirrors REST's. The PR body has a side-by-side alignment table against both `RESTServer._query` and `util:eval` so the choice of approach is auditable rather than ad-hoc.

**No eXide release needed** — once the merged existdb-openapi build is installed (next 0.9.x), the editor's behavior just becomes correct.

Two follow-ups falling out of this:

1. A performance PR here to add `XQueryPool` compiled-query reuse in `cursor:eval` (currently recompiles every call; REST doesn't).
2. ~~An upstream issue against eXist core: `RpcConnection.query(String xpath, ...)` over XML-RPC has the same defect class (the deprecated `queryP(path, ...)` is the variant that sets `STATIC_DOCUMENTS`). I'll file it separately so it can be tracked on its own.~~ (Update: Retracting this after determining it was off base. Empirical test against `/exist/xmlrpc` shows `query("count(//p)", ...)` returns the correct count — when `staticDocumentPaths` is null, `XQueryContext` defaults to walking all docs. `cursor:eval` was broken because `copyContext()` inherited a non-null `staticDocuments` cache from the caller, short-circuiting that default. Different bug; no upstream issue to file.)

line-o:
Waiting for a new release of existdb-openapi to test.

joewiz:
@duncdrum v0.9.4 is released:

- https://github.com/eXist-db/existdb-openapi/releases/tag/v0.9.4
- https://exist-db.org/exist/apps/public-repo/packages/existdb-openapi?eXist-db-min-version=7.0.0-SNAPSHOT

line-o:
I can confirm that count(//p) now returns the expected result! :tada:

line-o:
I still am able to trigger a state in Safari where I can no longer type into the editor window :/

I do see JSON-RPC events sent of received from the ws endpoint so my keyboard events to trigger but get swallowed

Screenshot 2026-05-30 at 15 57 54

line-o:
I opened http://localhost:8080/exist/apps/doc/modules/test-suite.xql

line-o:
Interesting: With that particular resource open I also cannot type in Chrome

line-o:
Outline is working

line-o:
It is either the size or the complexity of the resource that leads to this behaviour that one can no longer edit the contents of the file.

line-o:
Hmm, now I can edit the file in question.

line-o:
Run as test should check for dirty editor window ( as the changes to the testsuite have to be saved to the collection).

line-o:
I agree with @duncdrum that the error description is not enough. We need all the info (line, column, module ...) 
Maybe dump it in a first step and then we can ponder a better visualization afterwards.

line-o:
Screenshot 2026-05-30 at 16 11 56

or put the dump in a title attribute where we can still access it

joewiz:

@line-o — thanks for the rapid follow-up testing on v0.9.4 and the additional findings. Two of your three new items are addressed in the latest push to this PR; the third (Safari editor freeze) needs your help to diagnose.

## Run as test — dirty buffer + untitled (0fc8eef)

You're right: `api/test` reads from the stored collection via `source: doc.getPath()`, so unsaved edits were silently ignored. Two cases now handled:

- **Untitled buffer** (no path on disk): shows an actionable error toast — "Save the testsuite to the database before running its tests." — and skips the run.
- **Dirty saved buffer**: auto-saves first, then runs. If the save fails, surfaces the error rather than silently running stale code. Save success → run reflects what's on screen.
- **Clean buffer**: unchanged.

Regression test added: `cypress/e2e/run_as_test_dirty_spec.cy.js` covers both new branches.

## Error pill — full structured detail (b48d176)

Both you and @duncdrum asked for line / column / module / everything in addition to the description. The existdb-openapi error response already carries all of it (`{code, description, line, column, module, value}`); we just weren't surfacing it.

The pill stays compact, but now:

- **Pill `title` attribute** (hover) shows the full plain-text dump — every field on its own line. This is exactly your "put the dump in a title attribute where we can still access it" suggestion.
- **Detail panel** (click to expand) renders each field as a labelled row: `CODE`, `LOCATION`, `DESCRIPTION`, `MODULE`, `VALUE`. Fields that aren't present (e.g. synthetic `String/...` module IDs from inline-source compiles) are omitted to keep the panel readable.
- **Copy button** copies the full dump when structured data is available, the raw formatted message otherwise.

Regression test added: `cypress/e2e/query_error_structured_spec.cy.js` asserts the pill carries the title and the panel renders the fields.

Backwards-compatible: legacy callers that pass a plain string fall back to the original rendering path.

## Safari editor freeze — need a Performance profile

I couldn't reproduce this from code alone. The file you cited (`/exist/apps/doc/modules/test-suite.xql`) is only ~185 lines / 6 KB, so size isn't the explanation. Possibilities I traced from code reading:

- REx parser pathological case on the file's nested namespaced XML literals (`<db5:article>` + `xlink:` etc.) — `xquery-helper.js:parseXQuery` runs synchronously, blocking the main thread until done.
- If you have `autocompleteOnType: true` enabled in Preferences, every keystroke triggers another synchronous parse via `xquery-completion.js:78`.
- A blocking LSP/WS RPC — though your observation that "JSON-RPC events are sent/received" while typing is frozen makes this less likely.

The fact that **outline works** in the frozen state means the AST is built — so it's not an outright parser hang. The fact that it **sometimes self-resolves** suggests an async resource eventually unblocks.

Could you capture a Chrome (or Safari Web Inspector) **Performance profile** during the reproducer, ~10 seconds of recording covering: open the file → wait for outline → start typing → wait until input is swallowed → keep trying for a couple of seconds → stop recording? Attach the `.json` export here. That will pinpoint the blocking call to a specific function in milliseconds — much faster than me speculating. If profiling Safari is awkward, Chrome with the same reproducer would be equally diagnostic.

In the meantime, one cheap thing worth checking on your end: open Preferences and confirm whether "Autocomplete on type" is on or off. If on, try toggling it off and see whether the freeze still happens. That alone would narrow (a)/(b) vs (c).

line-o:
I am sometimes confused about the menu bar entries missing due to me not being logged in.

line-o:
The error display is OK but needs some tweaking (see screenshots)

is
Screenshot 2026-05-30 at 23 46 19

proposed
Screenshot 2026-05-30 at 23 49 25

```css
.ep-field-label {flex:1}
.ep-field-value {flex:6}
```

@duncdrum and I initially wanted to have the **test-case failure information**

Screenshot 2026-05-30 at 16 11 56

line-o:
The number of menu bar entries seems to be variable by some non-obvious mechanism. 
I would really want to see this being fixed list and _maybe_ disable entries that have no function in the current state.

Screenshot 2026-05-30 at 23 45 51
Screenshot 2026-05-30 at 23 46 03

line-o (approved):
It has quirks but I think we can improve on this in follow-up PRs

duncdrum:
Don't want to be a party p*** but isn't `7.0.0-beta` < `7.0.0-SNAPSHOT` so this won't install in the planned release? 

`7.0.0-alpha` should be safe

line-o:
According to the rules specified at https://semver.org

https://semver.org/#spec-item-9 states

> Pre-release versions have a lower precedence than the associated normal version.

That is _regardless_ of the pre-release version.

And later in https://semver.org/#spec-item-11 

> 4. Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:

>    1. Identifiers consisting of only digits are compared numerically.
>    2. Identifiers with letters or hyphens are compared lexically in ASCII sort order.
>    3. Numeric identifiers always have lower precedence than non-numeric identifiers.
>    4. A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.

> Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.

So because `SNAPSHOT` starts with an `S` it is therefore considered having a higher precedence than `beta` wich starts with a b.
But alpha is certainly not the solution. It would be - ironically - `rc.1` or `soap` or any pre-release identifier that comes later in the alphabet than `SNAPSHOT`.


-- update --

Since the implementation in semver.xq uses `lt` and `gt` to compare pre-release identifiers (if both are numeric or both are non-numeric) "rc.1" will not work. It must come after the capital letter S in ASCII so "RC.1"

```xquery
"beta" lt "SNAPSHOT",
"alpha" lt "SNAPSHOT",
"rc.1" lt "SNAPSHOT",
"RC.1" lt "SNAPSHOT"
```

line-o:
@duncdrum but this is the theory as we are still talking about the dependency resolution imposed by semver.xq

line-o:
I checked the implementation of semver.xq and it seems to follow the rules laid out by semver to the letter

https://github.com/eXist-db/semver.xq/blob/81b80ff4b547f4b95c9bd6e2b248d596249e9fcb/src/main/xquery/semver.xqm#L750-L768

duncdrum:
My point was that the processor version `semVer-min="7.0.0-alpha"` will install in both `7.0.0-beta` and `7.0.0-SNAPSHOT`

The change in 02d37eeb69fe04091c09f21533b7448d62944bb0 will not

The next dev version of exist should take `7.0.0-SNAPSHOT.1` if maven will let us, need to check

duncdrum:
after much soul-searching 
```xquery
semver:gt('7.0.0-beta', '7.0.0-SNAPSHOT')
```

is `true()` so we are good

joewiz:

Thanks for the approval, @line-o — and for the detailed follow-up notes from you and @duncdrum. And, I only noticed after posting this, *the merge* - thank you!

## Addressed in this PR

### CSS: error-pill value column widened (b81b6c1)

Took your proposed flex ratios (`label: 1` / `value: 6`) — pushed in b81b6c1. Field labels stay compact; values now get ~85% of the row.

### Run-as-test results — three real bugs fixed (c846b20)

Your "test-case failure information" screenshot prompted a fuller audit of the Run-as-test path. We had **zero** end-to-end Cypress coverage of the feature pre-this-PR, so when we wrote some, we found three real bugs in our own code (none caused by #794 — all latent):

1. **Server-side path bug** — `api/test.xqm` queried `$result/self::testsuite`, but `xqsuite:suite()` wraps results in `<testsuites><testsuite/></testsuites>`. The path never matched, so the JSON response had `time: ""` always (and `failures`/`errors` only worked because of a fallback). Fixed by using `$result/testsuite`.
2. **Missing `actual` value on failures** — XQSuite emits the expected value as the body of `<failure>` and the actual value as a sibling `<output>` element. We only exposed the expected half. The rendered detail said "expected: a" but you couldn't tell what the test actually returned. Server now returns both `expected` and `actual`; renderer formats them as labelled lines in the expandable detail.
3. **Unstructured error detail** — runtime errors now surface their `type:` separately in the detail panel.

New coverage in `cypress/e2e/run_as_test_results_spec.cy.js` (6 tests): all-pass, assertion failure, runtime error, mixed pass/fail/error, `%test:assertError`, and a regression test for the `time` field.

### semver-min — keeping `7.0.0-SNAPSHOT` (matrix confirms)

Ran `semver:satisfies-expath-package-dependency-versioning-attributes(...)` — the function EXPath actually uses to decide installability — against every realistic 7.0 candidate target value, for several candidate `semver-min` pins:

| `semver-min=` ↓ \ target → | `SNAPSHOT` | `SNAPSHOT.1` | `alpha` | `beta` | `rc.1` | `RC.1` | `7.0.0` |
|---|---|---|---|---|---|---|---|
| `7.0.0-alpha` | ❌ | ❌ | ✅ | ✅ | ✅ | ❌ | ✅ |
| **`7.0.0-SNAPSHOT` (current — 02d37ee)** | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ✅ |
| `7.0.0-RC` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `7.0.0` (no pre-release) | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ |

(All entries also `✅` for `7.0.1`, `8.0.0`.)

Key points the matrix surfaces:

- **`7.0.0-SNAPSHOT` covers everything except uppercase `RC.*`** because `'R'` (0x52) `<` `'S'` (0x53) in ASCII, and semver.xq follows [semver §11](https://semver.org/#spec-item-11) which mandates ASCII sort for identifiers with letters or hyphens. Lowercase `rc.1`, `alpha`, `beta`, plus `SNAPSHOT`/`SNAPSHOT.1` itself and the eventual `7.0.0` final, are all covered.
- **`7.0.0-alpha` would be worse, not better**: it doesn't cover `SNAPSHOT`/`SNAPSHOT.1` because `'S'` `<` `'a'` in ASCII, so the current dev builds wouldn't install. (My intuition for "alpha is the floor" was lexical-sort intuition; the actual semver rule is ASCII.)
- **The uppercase-`RC.*` row is a latent caveat, not a current blocker.** eXist historically tagged RCs as `5.0.0-RC8`, `5.0.0-RC7`, etc. (uppercase). If the 7.0 cycle reuses that convention we'd want to revisit, but `02d37ee` works against everything else and is easy to amend with a patch release if/when an uppercase-`RC.*` tag actually appears. A `semver-min="7.0.0-RC"` pin would close that gap pre-emptively but is semantically odd ("we require RC-or-newer").

So `02d37ee` stays as-is.

## Filed as follow-up issues

Three of your remaining points are best tracked separately rather than bolted onto this PR:

- **#805 — Run-as-test: further enhancements** (JUnit export, click-to-jump from test row to source line, pending-test display, setup/teardown failure rendering). The c846b20 fix above closes the immediate "show the failure detail" gap; this issue covers the broader UX.
- **#806 — Menu bar entries hide-vs-disable.** Both the logged-out missing entries and the "variable by some non-obvious mechanism" complaints. Proposed approach: disable rather than hide, plus a comment block documenting the conditional logic.
- **#807 — Safari editor freeze on certain XQuery files.** Reproducer needed; asks for a Performance profile to pinpoint the blocking call. Cheap pre-profile narrow: check whether Preferences → Autocomplete on type is on or off.

## Next

CSS, XQSuite, and semver-min are all addressed. The three follow-up issues capture the broader UX work.

PR #800, opened 2026-05-15
Update serialisation option to include media type

lguariento:
This fixes the

```
   XML Parsing Error: syntax error
   Location: https://site/exist/apps/eXide/outline
   Line Number 1, Column 1:
```

error in Firefox.

PR #799, opened 2026-05-15
Update JSON response handling in controller.xq

lguariento:
This addresses https://github.com/eXist-db/eXide/issues/253.

PR #801, opened 2026-05-21
Set the correct media type for login response and parse the Accept header 

### LINKED ISSUE #253: [BUG] XML Parsing Error in console
When using eXide, the following error keeps showing up in console, when logged in:

`XML Parsing Error: not well-formed Location: http://localhost:8080/exist/apps/eXide/login Line Number 1, Column 1:`

Inspecting that document, this shows up:

```
XML Parsing Error: not well-formed
Location: http://localhost:8080/exist/apps/eXide/login
Line Number 1, Column 1:
{ "user" : "admin", "isAdmin" : true }
^
```

eXist-db 5.2 or 5.3

marmoure:
This PR adds parsing for `Accept` header and introduces `local:accept-json` and `local:preferred-media-type`, it build on top of #800 to fix #253.
These functions allow more robust and future proof way for getting the appropriate header either the preferred one or a specific one.

Also when trying to set the correct headers for the login response using
```
util:declare-option("output:method", "json"),
util:declare-option("output:medita-type", "application/json"),
```
I encountered  a bug in the `RESTServer` in `writeResultJSON` which ignored the `output:medita-type` option, I added https://github.com/evolvedbinary/elemental/pull/212 for elemental and the same should be done for eXist-db.

In the mean time
```
response:set-header("Content-Type", "application/json; charset=UTF-8"),
``` 
is being used as  workaround until this bug is resolved.

### DIFF
--- a/controller.xq
+++ b/controller.xq
@@ -17,8 +17,6 @@ declare variable $local:config := config:get-configuration();
 declare variable $local:method := request:get-method() => lower-case();
 declare variable $local:uri := request:get-uri();
 declare variable $local:forwarded-for := request:get-header("X-Forwarded-URI");
-declare variable $local:wants-json := tokenize(request:get-header('Accept'), ', ?') = 'application/json';
-declare variable $local:is-ajax-request := request:get-header("X-Requested-With") = "XMLHttpRequest";
 
 declare function local:get-user () as xs:string? {
     let $login := login:set-user($local:login-domain, xs:dayTimeDuration("P7D"), false())
@@ -38,6 +36,34 @@ declare function local:query-execution-allowed($user as xs:string?, $is-dba as x
         $local:config/restrictions/@execute-query = "yes" and
         local:user-allowed($user)
     )
+};
+
+declare function local:parse-accept-header() as map(xs:float, xs:string+) {
+    let $media-type-maps as map(xs:float, xs:string)*  := (
+        let $media-types-and-quality := tokenize(request:get-header('Accept'), ',\s*')
+        for $media-type-and-quality in $media-types-and-quality
+        let $parts :=  tokenize($media-type-and-quality, ';\s*q=')
+        let $weight := if ($parts[2]) then xs:float($parts[2]) else xs:float(1.0)
+        let $media-type := $parts[1]
+        return
+            map { $weight: $media-type }
+     )
+     return
+            map:merge($media-type-maps, map { "duplicates": "combine"})
+};
+
+declare function local:preferred-media-type($media-type-map as map(xs:float, xs:string+) ) as xs:string {
+    let $highest-weight := sort(map:keys($media-type-map))[last()]
+    return
+        map:get($media-type-map, $highest-weight)[1]
+};
+
+declare function local:accept-json($media-type-map as map(xs:float, xs:string+)) as xs:boolean {
+    let $contains-json-media-type-results := map:for-each($media-type-map, function($key, $value) {
+        $value = "application/json"
+    })
+    return
+        $contains-json-media-type-results = true()
 };
 
 let $user := local:get-user()
@@ -118,9 +144,13 @@ else if (starts-with($exist:path, "/api/")) then
 
 else if (not($user-allowed))
 then (
-    if ($local:wants-json or $local:is-ajax-request)
+    if (local:accept-json(local:parse-accept-header()))
     then (
-        util:declare-option("exist:serialize", "method=json media-type=application/json"),
+        util:declare-option("output:method", "json"),
+        util:declare-option("output:media-type", "application/json"), (: does not work due to bug in writeResultJSON :)
+        (: the response:set-header("Content-Type" is used as a workaround until https://github.com/evolvedbinary/elemental/pull/212
+         : is merged on Elemental and the same should be done for eXist-db
+         Then the we can use the output:media-type option as expected :)
         response:set-header("Content-Type", "application/json; charset=UTF-8"),
         response:set-status-code(401),
         <status>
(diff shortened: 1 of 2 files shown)


duncdrum (changes requested):
Thanks for the PR @marmoure. The Xqdoc is not well-formed. 

We don't normally link elemental issues. Feel free to keep it in alongside an exist issue. But both should be properly formatted. 

Please add a test. Did you test this against exist? Which version?

marmoure:
Thanks for the @duncdrum I fixed the xqdoc issue, this was tested on eXist-db 6.4.1.

duncdrum:
@marmoure could you add a test for the request to the testsuite, so we see how this plays in against different backends.

adamretter:
@duncdrum I believe that @marmoure has mentioned in the past that the eXide test suite is flaky, but it doesn't look  like that has been addressed since then. Therefore we don't have much confidence in the eXide test suite.

We sent this code as a courtesy on behalf of one of your users of eXide. I am very reluctant to have @marmoure invest more time in this, because (a) we already have a fix for Elemental, and (b) previous attempts where we invested significant time into improving the eXist-db Apps in collaboration with you did not go well (i.e. https://github.com/eXist-db/monex/pull/298).

Can you provide any sort of assurance that if we invest further time and that if @marmoure provides a test for this, then you will merge it?

Thanks.

duncdrum:
@adamretter our review policy hasn't changed. You can see #799 for reference. Comparing this here to the full rewrite of monex is comparing apples to oranges.

What I can assure you is that I won't merge without a test. If you wish to invest more time into this version of eXide in light of #794  is up to you.

adamretter:
>  our review policy hasn't changed

> What I can assure you is that I won't merge without a test

@duncdrum Er... but... you just merged the following PR from another external contributor without a test - https://github.com/eXist-db/eXide/pull/799/changes

marmoure:
Hi @duncdrum I m bit confused here, I m seeing a CI fail and when I opened the screenshots I m confident that's not the code on this branch.
What is being tested in this case? and how can I reproduce this testcase locally?
with guest=yes (default) -- as guest user -- return login info as json (failed)

duncdrum:
@adamretter yes because the existing test suite caught a regression. As is visible in the comments. 

@marmoure  I m confused the screenshot is against latest major exide / exist v7. The issue was opened against previous / exist 6. 


What's the target? Rebase and fix against latest? Or still the original target.

marmoure:
> I m confused the screenshot is against latest major exide / exist v7. The issue was opened against previous / exist 6.

@duncdrum I m confused as well, I didn't expect to see screenshot from eXide V4 as this PR is quite old and still has the code for 3.5.5, the CI seems to be running the tests against V4.

> Rebase and fix against latest

Yes that would work, and I will do that, but I wanted to bring this matter to your attention first as I find it quite strange, if we are running the tests against the `Latest` exist and eXist-db does that mean only the tests are running? what about the code changes?

duncdrum:
@marmoure the code changes in this branch without rebase are v3. You should see that when you run the tests locally. 

What ci runs against depends on the ci script in this branch, and how exide is installed there.

marmoure:
Hi @duncdrum I rebased, and updated the test to work with the new V4 api.

duncdrum (commented):
Thank you @marmoure . The test is certainly helpful, I'm not sure if the controller change is still necessary in v4.

line-o:
@duncdrum the authentication will be removed entirely from the controller see #814

marmoure:
Thanks @duncdrum, What should I do in this case?

marmoure:
Hello, it's been a while can we get a verdict for this PR please.

joewiz:

@marmoure apologies — you asked "what should I do in this case?" on 17 June. Here is what I see.

**The controller changes are now moot, and the bug you set out to fix is gone.** #814 merged on 9 September and moved authentication out of `controller.xq` entirely — @line-o's June note here turned out to be exactly right. Login is now `POST /api/auth/session`, a Roaster route whose content type comes from the API definition, so no `util:declare-option` dance and no `response:set-header` workaround are needed.

Verified on current `develop` (eXide 4.1.0, `existdb/existdb:latest`, eXist-db 7.0.0-SNAPSHOT):

| request | result |
|---|---|
| `POST /eXide/login` — the URL in #253 | **404**; the route no longer exists |
| `POST /api/auth/session` | `200` · `Content-Type: application/json` |
| same, with `Accept: application/json` | `200` · `application/json` |
| bad credentials | `401` · `application/json` |

So #253 looks fixed to me — the console's "XML Parsing Error: not well-formed" came from that endpoint returning JSON under an XML content type, and both the endpoint and the mistyped response are gone. Worth closing it, and this PR with it.

**Your test, though, still earns its place, and it already passes.** I applied just the `auth_spec.cy.js` hunk to current `develop` and ran the spec: `return login info as json` passes, and the file stays green at 18/18. It guards precisely the regression class of #253 against the new API rather than the old controller, which is exactly what we'd want if anyone ever wires the login response up by hand again.

So the outcome I'd like is: drop the `controller.xq` commits, keep the test. Whichever is easier for you —

- you reduce this PR to the `auth_spec.cy.js` change and we merge it, or
- we cherry-pick that commit onto `develop` with you as author, and close this PR.

Happy to do the second so you don't have to spend more time here; just say which you prefer.

**On the `writeResultJSON` bug you found — it's fixed and released.** eXist-db/exist#6478 landed it on 2026-06-14 and it shipped in eXist-db 7.0.0-beta4 and 7.0.0-beta5. `writeResultJSON` now sets the response Content-Type before the output stream opens: `application/json` by default, and an explicit `output:media-type` is honored. Verified against a current 7.0.0-SNAPSHOT build:

```
POST /exist/rest/db   <query xmlns="http://exist.sourceforge.net/NS/exist" method="json">
  → Content-Type: application/json;charset=utf-8
```

Your report is what identified it, so thank you for that — it outlived this PR.

One thing that is still true, and is a different root cause rather than a leftover of the above: a stored `.xq` fetched over REST whose serialization method is declared *inside* the query (`declare option output:method "json"`) comes back as a JSON body under `application/xml`. The REST layer picks the JSON or XML writer by reading its own `output-as` property, while a query's own `declare option output:method` arrives under the W3C `method` key — so the JSON writer, and the Content-Type handling added in eXist-db/exist#6478, never run.

That is the shape eXide's old controller-forwarded login had, which is very likely what you were looking at on 6.4.1, so your report turned out to point at two adjacent defects rather than one. It is now fixed too: eXist-db/exist#6749, with a regression test that fails without it (`expected:<application/[json]> but was:<application/[xml]>`). BaseX 12.3 already behaves the way that PR makes eXist-db behave, for whatever that is worth as a cross-check.

Thanks for the patience, and sorry again for the silence.

adamretter:
@marmoure Could you as @joewiz suggests, remove the main code changes, and then just leave your tests. Hopefully then this can be merged and we can draw a line under it.

marmoure:
Hi @joewiz I updated the PR, it only adds the test to confirm https://github.com/eXist-db/eXide/issues/253 is fixed.

duncdrum (approved):

PR #812, opened 2026-06-05
fix(packages): the config endpoint applied configurations

line-o:
With this PR applied a request to
`/api/packages/{abbrev}/config` will now return with "Config Applied" if the configuration was valid and the collection was reindexed successfully.

Test for this needs to be added.


duncdrum:
@line-o I can confirm that this allows for proper reindexing with openapi 0.9.6 in local testing. But yes we need a test, this is a crucial operation we should guard against regressions.

joewiz:
@line-o What's the problem this PR solves? A test would perhaps clarify.

line-o:
I got a working test after some wrestling with cypress and the somewhat hard to test markup of eXide.

duncdrum (approved):

PR #814, opened 2026-06-06
fix(auth): move authentication to Roaster and retire the controller's auth logic

joewiz:

**Supersedes #813** — its commit is the first on this branch, so merging this delivers it too.

## Summary

Moves eXide's authentication wholly into Roaster and leaves `controller.xq` with no authorization logic at all. It began as the narrow "drop `security: []`, read `$request?user`" change stacked on #813 and grew, through @line-o's review, into the full migration.

## What it does

- **`controller.xq` carries no auth.** No persistent-login import, no `login:set-user`, no index-vs-login routing, no unauthorized redirect, no `sm:id()` read. It routes `/api/*` and the app shell to Roaster and otherwise serves static files.
- **Login and logout are ordinary Roaster routes.** `auth:login` calls `rauth:login-user` and `auth:logout` calls `rauth:logout-user`, both with the cookie name read from the spec — modelled on the roasted reference app @line-o pointed to.
- **The app shell is served by a route, `view:index` (`GET /index.html`), gated by `auth:is-allowed`.** A disallowed guest — under `restrictions/@guest = "no"` — is refused and redirected to `login.html`; the app HTML is never sent. Strict `guest=no` is enforced by Roaster rather than the controller.
- **`/execute` and `/results` are retired.** The editor runs queries through existdb-openapi's `/api/query` cursor, so both controller routes were dead; removing them deleted the controller's last authorization gate. `modules/{session,view,debuger,git}.xq` go with them.
- **"Remember me" now sets the session lifetime.** `duration` (a client-supplied string, never read) becomes a boolean `remember-me`; the server maps it to Roaster's `lifetime` — `P14D` remembered, `PT8H` not. `login.html` gains a real checkbox instead of an always-on hidden field.

## Safety

- **No CSRF behaviour change.** Neither eXide nor existdb-openapi declares `x-csrf`, so Roaster's `csrf:enforce` is inert — removing `security: []` from the login POST triggers nothing. Enabling CSRF protection remains a separate, deliberate step, and this makes the login POST a genuine cookie-auth route that it can protect when that happens.
- **One identity path for every route** — `$request?user`, populated by `rutil:getDBUser()` → `sm:id()` — matching existdb-openapi and the rest of the stack.
- **Requires Roaster >= 1.12.2**, already declared on `develop` via #869.

## Verification

On eXist-db 7.0.0-SNAPSHOT with the packages the image bundles:

- full Cypress suite green (257 tests); `auth_spec` covers the guest hard-refusal cases
- `Set-Cookie` `Max-Age` on `POST /api/auth/session`: 28800 (PT8H) with no flag, 1209600 (P14D) with `remember-me` set, over both form-encoded and JSON bodies
- guest under `guest=no` is redirected to `login.html` with no app HTML served; admin gets the app; query execution and paging work through the cursor

## Still open (not eXide)

For `guest=no` to fully close the editor's query path, existdb-openapi's `/api/query` needs to reject unauthenticated calls.


modules/api/auth.xqm
```diff
@@ -27,14 +30,20 @@ declare function auth:is-allowed($user as xs:string?) as xs:boolean {
...
+ :
+ : The persistent-login cookie is minted by controller.xq's `login:set-user`,
+ : which reads the `user` / `password` / `duration` form parameters before
+ : forwarding here. By handler time Roaster has resolved the resulting subject
+ : into `$request?user`; this handler reports it.
  :)
 declare function auth:login($request as map(*)) {
-    let $user := auth:get-user()
+    let $user := $request?user
+    let $name := ($user?name, "guest")[1]
```
line-o: I believe this fallback is neither necessary nor helpful

modules/api/auth.xqm
```diff
@@ -57,15 +66,15 @@ declare function auth:logout($request as map(*)) {
  : GET /api/auth/whoami — Current user info.
  :)
 declare function auth:whoami($request as map(*)) {
-    let $user := auth:get-user()
+    let $user := $request?user
+    let $name := ($user?name, "guest")[1]
```
line-o: see above

modules/api/auth.xqm
joewiz: [This response was co-authored with Claude Code. -Joe]

Agreed — removed in 6d1602a (both `auth:login` and `auth:whoami`).

You're right it's dead code: with `security: []` gone, these routes now run through `standard-authorization`, so `$request?user` is always populated — the guest account (`name` = "guest") when unauthenticated, as the module note says — and `$user?name` is never empty. The `("guest")[1]` branch could only ever fire on a genuinely-empty `$request?user`, which would be a regression worth surfacing rather than silently reporting "guest". Folding the routes back into the middleware is exactly what makes the fallback unnecessary, so good to drop it. Thanks for catching it.

line-o:
We need to start fresh:

- drop all authentication and login logic from controller.xq: this will all be handled by roaster
- make sure Roaster version 1.12.1 or later is declare as a dependency
- modify api.json to include the login route (if it does not already) for both form-based and XHR based logins
- in eXide modify the login route similar to the custom login handler in the roasted app (see https://github.com/eeditiones/roaster/blob/c9ef86bf52b51a3502a60d7e7a3e414e82bbfddb/test/app/modules/api.xql#L144-L161) to handle all logins

modules/api.json
```diff
@@ -40,6 +40,34 @@
...
+                            }
+                        },
+                        "application/x-www-form-urlencoded": {
+                            "schema": {
+                                "type": "object",
+                                "required": ["user", "password"],
+                                "properties": {
+                                    "user": { "type": "string" },
+                                    "password": { "type": "string", "format": "password" },
+                                    "duration": { "type": "string" }
```
line-o: Duration was added but is never read. I think this parameter should be dropped entirely. A boolean "remember-me" might be what we are after here.

modules/api/auth.xqm
```diff
@@ -31,35 +32,42 @@ declare function auth:is-allowed($user as xs:string?) as xs:boolean {
...
+ : username; on failure it returns the empty sequence.
  :)
 declare function auth:login($request as map(*)) {
-    let $user := $request?user
-    let $name := $user?name
+    let $user :=
+        rauth:login-user(
+            string($request?body?user),
+            string($request?body?password),
+            rauth:add-cookie-name($request, map {})
```
line-o: the empty map that is passed to `rauth:add-cookie-name` is the options map where we might want to set the duration for the session based if the user selected "remember-me" in the login form.

controller.xq
```diff
@@ -1,29 +1,30 @@
...
-    let $user-id := request:get-attribute($local:login-domain || ".user")
-    return $user-id
+(:~
+ : Authentication and login are handled entirely by Roaster on the /api/* routes
+ : (see modules/api/auth.xqm) — the controller carries no login logic. It still
+ : derives the current identity from sm:id() to authorize the one privileged
+ : non-Roaster route it serves, /execute.
+ :)
+declare function local:current-user() as xs:string? {
+    let $name := sm:id()//sm:real/sm:username/string()
```
line-o: `sm:id` might be called here _before_ cookie authentication took place.

joewiz:

Thanks @line-o — done, modeled on the roasted reference app. Pushed in `834036f`:

- **`controller.xq`** now carries no login logic: dropped the persistent-login import, `login:set-user`, the index-vs-login routing, the AJAX `login` route, and the unauthorized redirect/401 block. The page always loads and Roaster enforces access on the API — like roasted's controller.
- **Roaster `>= 1.12.1`** declared in `expath-pkg.xml(.tmpl)`.
- **`api.json`** — `POST /api/auth/session` now has `security: []` and a request body accepting both `application/json` (XHR) and `application/x-www-form-urlencoded` (the `login.html` form).
- **`auth.xqm`** — `auth:login` calls `auth:login-user(...)` and `auth:logout` calls `auth:logout-user(...)` (cookie name read from the spec via `auth:add-cookie-name`), following the roasted login handler you linked. Login still returns 401 for an account the config bars (guest when `restrictions/@guest = "no"`).

`login.html` now submits via `fetch` and redirects on success (the route returns JSON), and redirects already-authenticated visitors to `index.html`. The SPA's login modal already POSTed form-encoded credentials to `/api/auth/session`, so the frontend needed no change. `auth_spec` is updated for the new model (the app no longer gates the page server-side) and is green, along with the storage/query specs.

One thing I'd like your read on: `/execute` is the one privileged route that isn't a Roaster route (it forwards arbitrary XQuery to `XQueryServlet`), so I kept its authorization gate rather than drop it — re-sourced from `sm:id()` instead of the persistent-login attribute, returning a clean 403. Happy to move `/execute` under Roaster or drop the gate instead if you'd prefer; it just seemed safer not to leave arbitrary query execution ungated in this PR.

joewiz:

Pushed the `guest=no` fix and the auth-route cleanup — the controller is now fully auth-free. Summary for review:

**Guest gate — the hard refusal (`58da41a`)**

The app shell is now served by a Roaster route, `view:index` (`GET /index.html`), gated by `auth:is-allowed`. A disallowed guest — e.g. while `restrictions/@guest = "no"` — is refused and 302-redirected to `login.html`; the app HTML is never sent. `login.html` stays on an open route so login still works. So strict `guest=no` is back, but enforced by Roaster rather than the controller — per your suggestion to deliver the HTML via a route handler. The handler reproduces the former `view.xq` runtime-config injection and serializes as HTML5.

**Retired the legacy `/execute` + `/results` routes (`56dc3e0`, `afea6db`)**

The editor runs queries through existdb-openapi's `/api/query` cursor, so the controller's `/execute` (XQueryServlet) and `/results` (session paging) routes were dead. Removing them deleted the controller's **last authorization logic** — the `sm:id`-derived query-execution gate you flagged. The controller no longer imports `config`, reads `sm:id()`, or makes any auth decision; it routes `/api/*` and the app shell to Roaster and otherwise serves static files. Frontend result paging now uses the cursor exclusively (dropped `retrieveNext`). Also deleted the now-dead `modules/{session,view}.xq` and the already-retired (410) `modules/{debuger,git}.xq` (`debuger.xq` imported `session.xq`).

**Verified** on a Roaster 1.12.1 instance: guest under `guest=no` → 302 to login (no app HTML); admin → app loads; query execution + paging work via the cursor; auth/storage/query Cypress specs green (73/73). `auth_spec` gains the guest hard-refusal tests and drops the now-moot `/execute`-gate test.

**Still open (not eXide):** for `guest=no` to fully close the editor's query path, existdb-openapi's `/api/query` needs to reject unauthenticated calls. Working on this now in existdb-openapi.

**Merge ordering:** this branch is behind `develop`, which already has #825 (the `cy.execXQuery` test helper + the `cleanupTestFiles` fix) and #826. Best to bring #814 up to date with `develop` before merging it — at which point the dbmanager setup query here (currently an inline eXist REST `<query>` envelope) can switch to `cy.execXQuery`, and #825's `cleanupTestFiles` fix comes along for free.

duncdrum:
I think this needs a rebase

joewiz:

@duncdrum rebased onto current `develop` — the eight commits replayed with no conflicts, and the branch head is now `e73d059`.

One extra commit was needed to make the rebase actually work. This PR retires the `/execute` route, but `develop` has since grown a `cy.execXQuery` test helper (from #825) that posts to exactly that route, and `dbmanager_spec` calls it directly for setup. Straight after the rebase, every spec that seeds or cleans up state 404s. So the helper now posts an `<exist:query>` envelope to eXist-db's REST endpoint instead, `dbmanager_spec`'s inline call goes through the same helper, and the helper asserts a 2xx — posting *raw* XQuery to `/exist/rest` does not execute it (eXist-db parses the body as XML and 400s), a failure mode that used to pass silently because the caller ignored the status.

Verified on the stack the eXist-db image bundles (7.0.0-SNAPSHOT, existdb-openapi 0.10.0, Roaster 1.12.2): 278 tests, with the only failure being `query_error_structured`, which #840 fixes.

**The failing CI test here is not this PR.** Against the current image, the pinned `roaster-1.12.0` that CI installs cannot serialize eXide's login response:

```
400  Error while serializing xml: org.xml.sax.SAXException:
     err:SENR0001 Cannot serialize a map(*) with the XML or text output method
```

Every spec calls `cy.loginXHR` first, so the whole suite fails at spec 1 — on this branch, on #820, and on plain `develop` when reproduced locally. Swapping only the Roaster version to the bundled 1.12.2 makes the same request return 200. #865 fixes CI by deploying the packages the image bundles instead of pinned ones; this PR is additionally blocked on it, since it declares Roaster >= 1.12.1 and the pinned bed supplies 1.12.0.

So the order is #840, then #865, and this goes green.

duncdrum:
the quick fix is to allign the version in the workflow with the one specified in expath. See my response in the other PR

duncdrum:
@joewiz same here needs a rebase, could you also update the PR body, to reflect the status of the #813 discussion.

duncdrum:
@joewiz  needs another rebase onto current `develop` (diverged; conflicts with the Cypress  specs from #870).

While you’re in there, the Jun 16 review threads from @line-o are still open and should be resolved before merge:

1. **`duration` in the login form / OpenAPI schema** — still declared and sent from `login.html` (`P7D`), but never read. Drop it, or replace it with a boolean remember-me as suggested: https://github.com/eXist-db/eXide/pull/814#discussion_r3422996487  
2. **Wire remember-me into Roaster’s options map** — `rauth:add-cookie-name($request, map {})` currently passes an empty map; that’s where session duration should land if remember-me is kept: https://github.com/eXist-db/eXide/pull/814#discussion_r3423010635  

Also worth refreshing the PR body — it still describes the original “fold `security: []` / `$request?user`” stack on #813, but this branch has since seems to be the full Roaster auth migration.

joewiz:

@duncdrum rebased onto current `develop` (head `3e8dab1`), and @line-o's three June threads are addressed.

**The rebase conflict was the interesting part.** My branch still carried the pre-#870 `dbmanager_spec` and a `roaster semver-min="1.12.1"`. Both resolved in `develop`'s favour: #870's consolidated specs, and the `1.12.2` from #869 — 1.12.1 is the version measured to fail on login with `err:SENR0001`, so keeping this branch's value would have re-broken what #869 fixed.

**1. `duration` (https://github.com/eXist-db/eXide/pull/814#discussion_r3422996487)** — replaced with a boolean `remember-me`, as @line-o suggested, rather than dropped. It turned out not to be dead config: the SPA login dialog already offers "Remember me for 2 weeks" and `login.html` was sending a hidden `P7D` on every login, so the affordance was live and silently inert. Making it a boolean also closes the hole where any caller could name its own session lifetime — the client now says whether the box was ticked and the server decides what that means.

**2. The empty options map (https://github.com/eXist-db/eXide/pull/814#discussion_r3423010635)** — `auth:session-options` now builds the map that goes to `rauth:add-cookie-name`, setting Roaster's `lifetime`: `P14D` when remembered, `PT8H` when not. Both ends are set explicitly rather than leaving the unremembered case on Roaster's `P7D` default, which seems too long for a login that did not ask to be remembered — say the word if you would rather it stayed at the default.

`login.html` also gains a real checkbox, so the standalone page and the SPA now offer the same choice instead of the page quietly opting everyone in.

**3. `sm:id` before cookie authentication (controller.xq)** — resolved by the branch rather than by a fix: the commit retiring `/execute` and `/results` removed the last authorization logic from `controller.xq`, so there is no `sm:id` call left there at all.

Verified on eXist-db 7.0.0-SNAPSHOT with the packages the image bundles — `Set-Cookie` `Max-Age` on `POST /api/auth/session`:

| body | Max-Age |
|---|---|
| no flag | 28800 (PT8H) |
| form-encoded `remember-me=true` | 1209600 (P14D) |
| JSON `"remember-me": true` | 1209600 (P14D) |

Worth recording one trap for anyone wiring a similar flag: Roaster casts the value to `xs:boolean` when the body matches the schema, but a form-encoded body can arrive as the string `"true"`, and comparing the two directly is `err:XPTY0004`. The first cut of this returned a 500 for exactly that reason; the flag is now normalized through `string()`.

Full suite green (257 tests). The PR description is rewritten to describe the branch as it now stands — the full Roaster auth migration — rather than the original `security: []` change stacked on #813.

duncdrum (approved):

duncdrum:
ok confirmed that auth works locally, I found two unrelated minor bugs, on the way. PR will follow.

PR #826, opened 2026-06-16
fix(db): honor expand-xincludes on open

joewiz:

## Summary

Opening an XML document with `xi:include` returned the **expanded** content, and a subsequent save would write that expanded content back — silently dropping the `xi:include` directives from the source. A latent open→save data-loss bug.

## Verification

Serializing a document whose only child is `<xi:include href="target.xml"/>` (target = `<para>INCLUDED-CONTENT</para>`):

| serialization | result |
|---|---|
| eXist default (no setting) | `<para>INCLUDED-CONTENT</para>` — expanded |
| `util:declare-option("expand-xincludes=no")` (what develop ships) | `<para>INCLUDED-CONTENT</para>` — **still expanded (no effect)** |
| `fn:serialize` with `Q{…exist…}expand-xincludes: false()` | `<xi:include href="target.xml"/>` — preserved |

Verified live against the `/api/storage` open path: default open keeps `<xi:include>` (no data loss); `?expand-xincludes=true` expands to the included content.

Adds `cypress/e2e/xinclude_serialization_spec.cy.js` guarding both directions. Mutation-verified: reverting `db:load-document` to the `util:declare-option` approach fails the "default preserves xi:include" assertion while the expand assertion still passes — so the guard demonstrably guards.

## Scope / notes

- Test-path only changes one branch of `db:load-document` (non-binary, non-download). Binary and download paths are untouched.
- No upstream dependency; independent of the db-core adapter work (#824). When the open path later cuts over to existdb-openapi's `/api/db/resource`, that endpoint honors `expand-xincludes` server-side, so eXide can pass `?expand-xincludes=no` and drop its own serialization handling entirely — at which point this branch of `db:load-document` goes away.

### DIFF
--- a/cypress/e2e/xinclude_serialization_spec.cy.js
+++ b/cypress/e2e/xinclude_serialization_spec.cy.js
@@ -44,4 +44,20 @@ context('XInclude serialization on open', () => {
       expect(r.body.content).to.not.contain('<xi:include')
     })
   })
+
+  // Anticipates the cutover described in the PR's Scope/notes: once the open path moves
+  // to existdb-openapi's /api/db/resource, db-core honors expand-xincludes server-side
+  // and db:load-document's own serialization handling goes away. This asserts that
+  // endpoint's contract directly. Skipped until the cutover lands -- it needs the
+  // consolidated db-core (existdb-openapi #55/#59), which is not in eXide's current
+  // test environment.
+  it.skip('future: existdb-openapi /api/db/resource honors expand-xincludes server-side', () => {
+    const path = encodeURIComponent('/db/cypress-xinclude/main.xml')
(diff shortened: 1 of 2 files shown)


duncdrum (commented):
@joewiz could you add a skipped test for the form that we expect to work in the future?

duncdrum (approved):

joewiz:
@duncdrum Thanks! 

p.s. Before committing Claude smoke-verified the contract against a current eXist-db/existdb-openapi#59 build and confirmed that `?path=…&expand-xincludes=false` returns the document with `<xi:include>` preserved, and `…=true` expands it — which is what the skipped test asserts, so it should flip green once #824 can be merged and we enable this test.

PR #865, opened 2026-08-19
ci: test both the declared dependencies and eXist-db's bundled ones

joewiz:

## Problem

The workflow started the test container with:

```yaml
docker run --volume $(pwd)/build:/exist/autodeploy:ro ... existdb/existdb:latest
```

A bind mount *replaces* the directory, so everything the eXist-db image ships in `/exist/autodeploy` was discarded and CI deployed exactly three packages: the freshly built eXide, plus the two the workflow downloaded — `roaster-1.12.0` and `existdb-openapi-0.9.5`.

That is not the stack anyone runs. `existdb/existdb:latest` bundles:

```
dashboard-2.0.9.xar                      monex-5.1.0.xar
eXide-4.0.1.xar                          packageservice-1.3.14.xar
exist-documentation-7.0.0.xar            roaster-1.12.2.xar
exist-function-documentation-3.0.1.xar   semver-xq-3.0.0.xar
existdb-openapi-0.10.0.xar               templating-1.2.1.xar
functx-1.0.1.xar
```

eXide's own README tells contributors to `docker cp ./build/*.xar exist-ci:exist/autodeploy`, which *layers* onto that set — so the documented local procedure and CI have been testing different stacks.

## The pins are now actively broken

As of today's `existdb/existdb:latest`, the pinned `roaster-1.12.0` cannot serialize eXide's login response, so `POST /api/auth/session` returns:

```
400  Error while serializing xml: org.xml.sax.SAXException:
     err:SENR0001 Cannot serialize a map(*) with the XML or text output method
```

Every spec calls `cy.loginXHR` first, so the whole suite fails — 100% of specs, on any branch. Isolated locally against the same image, with only the Roaster version changed:

| eXide | Roaster | existdb-openapi | `POST /api/auth/session` |
|---|---|---|---|
| `develop` | 1.12.0 (the pin) | 0.9.5 (the pin) | **400** |
| `develop` | 1.12.2 (bundled) | 0.9.5 (the pin) | 200 |
| `develop` | 1.12.2 (bundled) | 0.10.0 (bundled) | 200 |

`develop`'s last green run predates this; the next push to `develop` goes red. This PR is the fix.

## Why the divergence mattered before that

This is also what made eXide#821 unresolvable for two months. The pinned `existdb-openapi-0.9.5` predates the `/api/query` error-envelope regression, so CI stayed green while every user on a stock eXist-db 7 hit a query error eXide could not display. Reproduced on a clean container, same eXist-db build, same query:

| existdb-openapi | `/api/query` response to `1 + "oops"` | query-error specs |
|---|---|---|
| 0.9.5 (the pin) | `HTTP 500` + `{ code, line, column, description }` | pass |
| 0.9.7 (shipped with 7.0.0-beta3) | `HTTP 200` + `{ error: "Invalid context-item: …" }` | fail |
| 0.10.0 (current) | `HTTP 400` + `{ code, message, line, column, raw }` | fail without #840 |

The pins also block PRs: eXide#814 requires Roaster >= 1.12.1 and cannot pass on a bed that supplies 1.12.0.

## Change

Assemble the autodeploy set from the image's own packages, with only eXide swapped for the build under test:

```yaml
CID=$(docker create existdb/existdb:latest)
docker cp "$CID":/exist/autodeploy/. autodeploy/
rm -f autodeploy/eXide-*.xar
cp build/eXide-*.xar autodeploy/
```

The two `gh release download` steps go away with the pins. Copying the set out and mounting it (rather than `docker cp`ing the xar in) keeps the deployment deterministic: exactly one eXide, even across a version bump.

## Merge ordering

CI is red right now regardless of this PR, so this is about the shortest path back to green rather than about avoiding a regression.

**Merge eXide#840 first, then this.** Against the bundled existdb-openapi 0.10.0, `query_error_structured_spec` fails on current `develop` — the real incompatibility this PR stops hiding, and #840 is its fix. With #840 in, this PR takes the suite to fully green on the stack users actually run.

## Verification

Full 39-spec suite, clean container each run, `existdb/existdb:latest` (7.0.0-SNAPSHOT, build 2026-08-18):

- `develop` + bundled packages (openapi 0.10.0): 1 failing — `query_error_structured`, per the ordering note above
- `develop` + eXide#840 + bundled packages: query-error specs pass
- `develop` + the old CI pins (roaster 1.12.0, openapi 0.9.5, no other apps): **all 38 specs fail at login**, per the table above

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,12 @@ jobs:
          exist-version: [latest]
          # Node 22 is EOL; track current Node LTS (24).
          node-version: ['24']
+         # Two beds, two questions (eXide#821, eXide#865):
+         #   declared — do the dependency versions declared in expath-pkg.xml
+         #              actually run the app? Nothing else is installed.
+         #   bundled  — does the app work with what eXist-db ships by default?
+         #              The build under test is installed over the bundled copy.
+         deps: [declared, bundled]
 
     steps:
       - uses: actions/checkout@v7
@@ -42,18 +48,37 @@ jobs:
       - name: Build frontend Using node.js ${{ matrix.node-version }} and build EXPath Package
         run: npm run build
 
-      - name: Download Roaster dependency
-        run: gh release download v1.12.0 --repo eeditiones/roaster --pattern "roaster-1.12.0.xar" --output build/roaster-1.12.0.xar
+      # The versions come from expath-pkg.xml rather than being hardcoded here,
+      # so this bed tests the contract eXide publishes to the Package Manager.
+      # If a declared minimum is wrong, this leg fails and the declaration is
+      # what needs fixing.
+      - name: Resolve the declared dependencies
+        if: matrix.deps == 'declared'
+        id: declared
+        run: |
+          # Reads the semver-min values out of expath-pkg.xml; fails if any
+          # dependency declares none. Node, not xmllint — the runner image does
+          # not ship libxml2-utils.
+          eval "$(node tools/read-declared-deps.js)"
+          echo "Declared minimums: roaster $ROASTER, existdb-openapi $OPENAPI"
+          gh release download "v$ROASTER" --repo eeditiones/roaster \
+            --pattern "roaster-$ROASTER.xar" --output "build/roaster-$ROASTER.xar"
+          gh release download "v$OPENAPI" --repo eXist-db/existdb-openapi \
+            --pattern "existdb-openapi-*.xar" --dir build/
         env:
           GH_TOKEN: ${{ github.token }}
 
-      - name: Download existdb-openapi dependency
-        run: gh release download v0.9.5 --repo eXist-db/existdb-openapi --pattern "existdb-openapi-*.xar" --dir build/
-        env:
-          GH_TOKEN: ${{ github.token }}
-
-      - name: Start eXist-db Docker Container
+      # declared: only the built xar and its declared dependencies are present.
+      - name: Start eXist-db with the declared dependencies
+        if: matrix.deps == 'declared'
         run: docker run --rm --name exist --volume $(pwd)/build:/exist/autodeploy:ro --publish 8080:8080 --detach existdb/existdb:${{ matrix.exist-version }}
+
+      # bundled: stock image, nothing removed or extracted; the freshly built
+      # xar is installed over the bundled eXide once the server is up, so the
+      # app under test is unambiguously the one this run compiled.
+      - name: Start eXist-db with its own bundled packages
+        if: matrix.deps == 'bundled'
+        run: docker run --rm --name exist --publish 8080:8080 --detach existdb/existdb:${{ matrix.exist-version }}
 
       - name: Wait for eXist-db Startup and Package Deployment
         run: |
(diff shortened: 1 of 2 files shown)


duncdrum:
@joewiz yes this is a know problem of all our Ci configs. Luckily that means there is plenty of prior art for how to approach this. 

The basics principles are:

1. expath dependencies are declared in `expath-pkg.xml` CI results must show that the deps declared there are the ones the app runs correctly with. (If it doesn't they have to be adjusted)
2. A second matrix should test with whatever ships by default with exist, as that's of most relevance to our users.

Using a volume with just the deps declared in expath-pkg ensures 1. But requires determining deps on CI either manually (what we have here) or by parsing expath-pkg.xml inside the workflow.

For 2. one can just drop the whole volume from CI, but then one has to ensure that the app under test is actually the xar being compiled, and not the one inherited from the base image.

using `-slim` images and `xst install` is another approach for small and medium apps. 

None of these solutions are great, they are all hacks around the shortcomings of auto deploys and its lack of dependency resolution. 

Extracting the xars from the image is not the right approach imv, it fails to satisfy 1. and it's just a long-winded way to tackle 2.

joewiz:

@duncdrum this is a better frame than my original approach, and you are right that extracting the xars conflates your two principles. Reworked along your lines — the branch is now a `deps: [declared, bundled]` matrix.

**declared** resolves the versions *from* `expath-pkg.xml` rather than hardcoding them here, and installs only those. If a declared minimum is wrong the leg fails, which is the behavior you asked for. **bundled** starts the stock image with nothing removed, then installs the freshly built xar over the bundled copy with `xst package install --force`, which is a cleaner answer to your "ensure the app under test is actually the xar being compiled" than anything involving the volume.

Two findings from implementing it that are worth having on the record.

**Principle 1 currently fails, and the declarations are why.** eXide declares:

```xml
<dependency package="http://e-editiones.org/roaster" semver-min="1.8.0"/>
<dependency package="http://exist-db.org/pkg/openapi"/>
```

Roaster's declared floor is off by four minor versions, and existdb-openapi has no minimum at all. Varying only the Roaster version against eXist-db 7.0.0-SNAPSHOT — `POST /api/auth/session`:

| Roaster | result |
|---|---|
| 1.12.0 | 400 — `err:SENR0001 Cannot serialize a map(*) with the XML or text output method` |
| 1.12.1 | 400 — same |
| 1.12.2 | 200 |
| 1.13.0 | 200 |

So the real minimum is 1.12.2 — worth noting for #814 as well, where 1.12.1 was the figure under discussion. I have opened a separate PR correcting both declarations, #869 (Roaster 1.12.2, existdb-openapi 0.10.0); **it needs to land before this one**, because the declared leg deliberately fails while no `semver-min` is declared for existdb-openapi.

**No `-slim` tags exist.** `existdb/existdb` publishes 60 tags and none of them are slim variants, so that route is not available today — happy to switch the declared leg to it if slim images are planned, since `xst install` is already how the bundled leg works.

Verified locally on eXist-db 7.0.0-SNAPSHOT, both legs: 261 tests, 257 passing, 3 pending, the only failure being `query_error_structured`, which #840 fixes. On the bundled leg the installed descriptor reports this build's declared minimums rather than the bundled package's, confirming the built xar is the one under test — and CI's own bundled leg agrees: `installed eXide@4.0.1` from the build, then 261 tests with `query_error_structured` plus one autocomplete flake (the same one #866 is about, now seen twice on CI and never locally).

The declared leg needs the declaration PR before it can go green, and its first CI run also caught an implementation detail worth flagging for the other repos: `xmllint` is not on the `ubuntu-latest` runner any more, so the parse now runs through a small Node script (`tools/read-declared-deps.js`) that reads the `semver-min` values and fails loudly when a dependency declares none.

duncdrum (approved):

duncdrum:
Needs a rebase

joewiz:
@duncdrum Rebased. Once this is merged, the one remaining blocker on a release is a review on #840.

duncdrum (approved):

Issue #821, opened 2026-06-11
Failing tests against eXist-db/Elemental


marmoure:
We have some users reporting errors with the latest version of eXide. We wanted to investigate and report those errors, but first discovered that there are already failing tests — so we felt we should report those first so you can fix them and establish a stable baseline for testing.

I tested against eXist-db 7-Beta-3 and Elemental 7.6.0 and found failing tests on both. Here is a summary:

| Test | eXist-db | Elemental |
|------|----------|-----------|
| Query error display - type error (XPTY0004) | Pass | Fail |
| Query error display - structured fields exposes structured fields in pill title and panel body | Pass | Fail |
| Run as test - results rendering clicking a test row jumps to the annotation line in the editor | Pass | Fail |
| WebSocket transport - connects to /exist/ws endpoint | Pass | Fail |
| WebSocket transport - handles ping messages without errors | Pass | Fail |
| WebSocket transport - receives monitoring data via WebSocket push | Pass | Fail |
| WebSocket transport - receives diagnostics push after compilation | Pass | Fail |
| Apply configuration - saves the configuration and shows the apply configuration dialog | Fail | Pass |
| DB Manager - DB Manager operations - create - should open resource | Fail | Pass |
| Monitor panel - starts polling when toggled open by admin | Fail | Pass |
| Monitor panel - updates memory bar on poll | Fail | Pass |
| Monitor panel - shows uptime after polling | Fail | Pass |
| Monitor panel - stops polling when toggled closed | Fail | Pass |
| Monitor panel - shows running query and allows killing it | Fail | Pass |
| Monitor panel - restores monitor on reload if it was open | Fail | Pass |
| Error status UI - copy button copies error text to clipboard | Fail | Pass |

### Elemental 7.6.0

1. query_error_display_spec.cy.js
2. query_error_structured_spec.cy.js
3. run_as_test_results_spec.cy.js
4. websocket_spec.cy.js

### eXist-db 7-Beta-3

1. dbmanager_spec.cy.js
2. monitor_spec.cy.js
3. apply_configuration.cy.js
4. lsp/error_status_spec.cy.js

The full log from the eXist-db 7-Beta-3 run is attached.

## Environment

Java: 21.0.5 lts temurin

OS: Linux Mint 22.3

eXide: [4.0.1](https://github.com/eXist-db/eXide/releases/tag/v4.0.1)

[...]

PR #866, opened 2026-08-19
test(cypress): raise the default timeout and enable retries

joewiz:

## Problem

Several assertions wait on a server round trip while using Cypress's 4-second default (or a hand-set 5 seconds). That is comfortable on a warm instance and tight on a loaded CI runner or a freshly started container, so these specs fail intermittently and are then hard to distinguish from real regressions — which is exactly what made the reports in #821 so hard to act on.

Three sightings, all of which passed on an immediate re-run:

| spec | assertion | where |
|---|---|---|
| `autocomplete_spec` | `.cm-tooltip-autocomplete` after "shows repo: completions when typing repo without colon" | #865's first CI run |
| `apply_configuration` | save dialog closed, first save against a cold container | local full-suite run |
| `websocket_spec` | diagnostics push after compile | local full-suite run |

The autocomplete one also appears in the failure tables reported in #821.

## Change

Raise the timeout only on assertions that are waiting on the server, and say so in a comment where it is not obvious:

- **10s** — the completion popup (served by the language service), the save dialog opening and closing, the "stored" / "Configuration applied." toasts, and the path label after a save
- **15s** — `websocket_spec`'s diagnostics assertion, which waits for a compile request *and* a push back over the socket

Assertions that only wait on local DOM work are left at their existing values; this is not a blanket bump.

## Verification

Full suite on the stack the eXist-db image bundles (eXist-db 7.0.0-SNAPSHOT, existdb-openapi 0.10.0, Roaster 1.12.2), clean container: the three specs above pass, and the suite is otherwise unchanged from `develop`.

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -4,6 +4,11 @@ module.exports = defineConfig({
   projectId: '6tyg6v',
   viewportWidth: 1200,
   viewportHeight: 860,
+  // Assertions that wait on a server round trip need more than Cypress's 4s
+  // default against a cold or loaded instance; retries absorb the rest of the
+  // timing noise without hand-tuning individual specs.
+  defaultCommandTimeout: 10000,
+  retries: { runMode: 2, openMode: 0 },
   e2e: {
     // We've imported your old cypress plugins here.
     // You may want to clean this up later by importing these.


duncdrum (commented):
There are two leaner option: 
- adjust the default timeouts in cypress config, instead of per testcase
- configure retries in cypress configs (this works really well in my experience) to get rid of more flake

The third option (most work but best results):
Where possible make longer chains using cy.request with assertions on outgoing and request and incoming responses, where each one starts a new timeout window, and failures become much more informative

joewiz:

Thanks @duncdrum! I tried your first two suggestions together, and they do the job with a much smaller diff, so I am happy to switch this PR to them:

```js
defaultCommandTimeout: 10000,
retries: { runMode: 2, openMode: 0 },
```

With the per-case edits reverted, that gives 261 tests, 257 passing, and the one genuine failure (`query_error_structured`, which #840 fixes) still failing after its retries — so retries absorb the noise without hiding the real thing.

Two observations before I make the swap, since you have more experience with retries here than I do.

**Retries and specs that chain state.** Several specs carry state across tests rather than being independent: `file_save_spec`'s third and fifth tests operate on the file its second test saves, and `dbmanager_spec` runs create → rename → delete sequences. A retried test in those chains re-runs against state its earlier siblings already changed, so it can pass or fail for reasons unrelated to the original failure. That is survivable, but it argues for making those specs independent at some point rather than relying on retries to paper over them.

**Cost of real failures.** The genuinely failing spec took 31s instead of 5s, since it now fails three times. Fine at this suite size; worth knowing before the suite grows.

Your third option is clearly right for `websocket_spec`'s diagnostics assertion specifically — it waits on a compile request *and* a push back over the socket, and a `cy.request` chain with assertions on each step would both give it its own timeout window and say which half was slow. I would rather do that as a follow-up than fold it into this PR.

Two data points since I opened this: the autocomplete tooltip has now failed twice on CI (both times on #865's runs, where the suite actually gets past login) and never once locally, which is about as clear a load-sensitivity signal as we are going to get.

Note this PR's failing CI test is the pinned-Roaster problem from #865, not anything here.

duncdrum:
I m afraid the preferred design is to not have state shared between testcases. Long chains with multiple `cy.get` and assertions are the cleaner option combined with cleanup via `after` or `afterEach` 

So I m not sure how retry behaves with shared state. 

And yes isolated specs for shared state stuff is a good idea.

joewiz:

@duncdrum agreed on the design, and your uncertainty about retries plus shared state matches mine — that is why I raised it rather than just turning retries on. Cypress retries the failing test only, not the ones that set up its state, so in a chained spec a retry re-runs against whatever the earlier tests left behind. It can go either way and neither outcome is trustworthy.

Where the suite stands on that today:

| spec | tests | shared state between tests |
|---|---|---|
| `dbmanager_spec` | 22 | create → rename → delete sequences |
| `file_save_spec` | 5 | tests 3 and 5 operate on the file test 2 saves |
| `apply_configuration` | 3 | saves a file, then a config for it, then applies it |

Proposal: keep this PR to the config change so it can go in and stop the noise, and do the independence work as a follow-up — each test seeding what it needs and cleaning up in `after`/`afterEach`, with the long `cy.get` chains and assertions you describe. `file_save_spec` and `apply_configuration` are small and would come first; `dbmanager_spec` is the real job, since several of its sequences are deliberately testing a UI workflow and want restructuring into self-contained chains rather than a mechanical split.

One caveat I would rather state than discover later: until that work lands, retries are doing exactly the ambiguous thing you describe on those three specs. The trade seems worth it — two CI sightings of the autocomplete flake and no local reproduction — but it is a trade, not a fix.

duncdrum (approved):

PR #870, opened 2026-08-30
test(cypress): make specs independent of one another

joewiz:

## Why

Follow-up to @duncdrum's review on #866: the preferred design is no state shared between test cases. It also matters mechanically now that retries are on — Cypress retries only the failing test, not the ones that built its fixture, so a retried test in a chained spec runs against whatever its predecessors left behind and neither outcome tells you anything.

Three specs carried that shape:

- **`file_save_spec`** — three tests operated on the file a fourth test saved
- **`apply_configuration`** — one collection created once for the whole spec, with each test writing into it, and the third test opening the configuration the second one saved
- **`dbmanager_spec`** — every describe block was one workflow split across several `it`s ("create", then "rename", then "delete"), and the copy and cut blocks shared the same two collection names, so a failure in copy's cleanup broke cut

## Change

Every test now owns its state. Each names its own file or collection (unique per test, under a prefix the cleanup sweeps), seeds what it needs server-side rather than through a sibling test's UI flow, and cleans up in `afterEach`. In `dbmanager_spec` each test drives one workflow end to end with assertions along the way, which turns 22 tests into 9 — none of which depend on another.

Three assertions got sharper in the process, all cases where the old split had left a gap:

- the save test asserts the resource is really in the database via `GET /api/storage`, rather than trusting the editor's own path label
- copy and cut previously differed only in which toolbar button they clicked, and neither checked what became of the source; they now assert that a copy leaves the original in place and a cut does not
- the rename test asserts the old name is gone, not just that the new one appeared

## Verification

Against eXist-db 7.0.0-SNAPSHOT with the packages the image bundles:

- all three specs pass as a whole
- **each of their seventeen tests passes when run on its own** (`it.only`, one at a time) — which is the property this PR is actually about
- the database is left with no `cypress-*` collections afterwards
- the full suite still reports only `query_error_structured` failing, which #840 fixes (the total drops from 261 to 248 because `dbmanager_spec`'s 22 tests become 9)

## Note on the test count

The suite reports fewer tests than before (248, down from 261) because `dbmanager_spec`'s create/act/delete sequences are now one test each rather than three. Coverage is the same or slightly better — the steps that used to be separate `it`s are still executed and still asserted, now inside the chain that depends on them.

### DIFF
--- a/cypress/e2e/apply_configuration.cy.js
+++ b/cypress/e2e/apply_configuration.cy.js
@@ -1,6 +1,13 @@
+/**
+ * Each test owns its collection: it is created before the test and removed
+ * after, and any file a test needs is seeded rather than inherited from the
+ * test before it. With retries enabled (#866) Cypress re-runs only the failing
+ * test, not the ones that built its fixture, so shared state makes a retry
+ * meaningless.
+ */
 describe('Apply configuration', () => {
-  // Use /db which always exists; clean up test files after
-  const testCollection = '/db/test-collection-' + Date.now()
+  // Assigned per test in beforeEach, so no two tests share a collection.
+  let testCollection
   const testFile = 'testcontent.xml'
   const testConfigFile = 'collection.xconf'
   const testConfig = `<collection xmlns="http://exist-db.org/collection-config/1.0">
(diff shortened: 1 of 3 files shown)


duncdrum (approved):

duncdrum:
@joewiz needs a rebase, and review of CI results.

PR #815, opened 2026-06-08
ci: set up semantic-release

line-o:

PR #875, opened 2026-09-07
feat(release): automated releases via exist-bot, mirroring monex#455/#456

duncdrum:
## What

Rebases #815's semantic-release setup onto current develop and brings eXide's release automation up to the pattern established in eXist-db/monex#455 and #456.

## Why

#815 was branched before a lot of develop's history landed and needed a rebase (per review comment). Rebasing it as-is surfaced three problems that would have made the release pipeline either fight with existing CI or silently never fire/publish a correct version:

1. **Duplicate CI workflow.** #815 added `test.yml` as a copy of `ci.yml` made before `ci.yml` grew its declared/bundled dependency matrix (eXide#821, eXide#865). Both would run on every push/PR, and the release workflow would watch the stale copy.
2. **Wrong branch.** `.releaserc` and `release.yml` targeted `main`, which doesn't exist on this repo — the default/integration branch is `develop`.
3. **Wrong build order.** `npm run build` ran *before* semantic-release determined the next version, so the XAR's embedded `expath-pkg.xml`/`index.html` version (and `package.json`) would always stay `0.0.0-development` regardless of the GitHub release tag.

## How

- Removed the duplicate `test.yml`; `release.yml` now triggers off `ci.yml`'s actual `exist-db CI` workflow_run, on `develop`.
- Added `@semantic-release/exec` (`prepareCmd` bumps `package.json` via `npm version` before `publishCmd` runs `npm run build`) and `@semantic-release/git` (commits the version bump back), matching monex's `.releaserc`.
- **Release job now uses the exist-bot GitHub App token** (`actions/create-github-app-token`) instead of the default `GITHUB_TOKEN`, for the same reason as monex#455: the default token can't push the `chore(release)` commit through branch protection on `develop`, and the built-in `github-actions` identity isn't installable as a ruleset bypass actor. Reuses the org-level `RELEASE_BOT_APP_ID`/`RELEASE_BOT_PRIVATE_KEY` already set up for monex.
- **New `publish-to-public-repo.yml`**, copied from monex#456: on `release: published`, downloads the XAR asset and POSTs it to `https://exist-db.org/exist/apps/public-repo`, using the org-level `PUBLIC_REPO_USERNAME`/`PUBLIC_REPO_PASSWORD`. Includes a `workflow_dispatch` retry/backfill path.

## Before merging

This repo needs the same one-time GitHub-side setup monex already has:
- Install the exist-bot App on `eXist-db/eXide` with **Contents: read & write**.
- Add exist-bot as a bypass actor on `develop`'s branch protection/ruleset.

(`RELEASE_BOT_APP_ID`, `RELEASE_BOT_PRIVATE_KEY`, `PUBLIC_REPO_USERNAME`, `PUBLIC_REPO_PASSWORD` are already org-level, so no new secrets are needed.)

## Verification after merge

1. Merge a `fix:`/`feat:` commit to `develop` and confirm the release job pushes as `exist-bot[bot]`, tags, and creates a GitHub Release with a correctly-versioned XAR attached.
2. Confirm `Publish to public-repo` fires on that release and https://exist-db.org/exist/apps/public-repo/packages/eXide lists the new version.

Closes #815.

https://claude.ai/code/session_01DZJC3w5WHeiTCHZ2JqrJJw


duncdrum:
Pushed a follow-up fix from code review: `.husky/commit-msg` and the `prepare` script were wired up with husky's removed v8 API (`husky install` / sourcing `_/husky.sh`), which only worked via a deprecated compat shim husky says will be dropped in v10 — a routine Dependabot bump would have silently broken every local commit. Migrated both to the v9+ format and verified the hook passes/fails correctly locally.

One remaining known limitation, left as-is rather than "fixed": `package.json`'s `version` only gets bumped to a real value during the release job (via `.releaserc`'s `@semantic-release/exec` `prepareCmd`). Any other build — a developer running `npm run build` locally, or the `declared`/`bundled` matrix jobs in `ci.yml` — still produces `build/eXide-0.0.0-development.xar` and an `expath-pkg.xml` declaring that placeholder version. That's cosmetic for CI/local testing (nothing consumes that version outside the release path) but worth knowing if it shows up in a build artifact.

duncdrum:
Pushed one more piece, closing a gap in the original scope: the pipeline generated GitHub Release notes but never touched `repo.xml` — the descriptor eXist-db's Package Manager actually shows users browsing/installing the app — so its `<changelog>` would've gone stale the moment releases stopped being hand-cut.

Ported the approach monex uses for this: `repo.xml` is now a template (`repo.xml.tmpl`, same idiom as the existing `expath-pkg.xml.tmpl`/`index.html.tmpl`), and a new `prepareCmd` step (`scripts/update-repo-changelog.mjs`) parses conventional commits since the previous tag and inserts a `<change version="X">` entry before the build runs. `@semantic-release/git` now also commits `repo.xml.tmpl` back alongside the version bump.

Heads up for whoever reviews: this switches `repo.xml`'s changelog style going forward from eXide's current hand-written prose entries (e.g. the 4.0.0 "Modernization release" summary) to mechanical one-line-per-commit entries grouped New/Fix/Improvement — same tradeoff monex accepted. There's an in-progress npm package, [`@existdb/repo-xml-changelog-generator`](https://github.com/eXist-db/repo-xml-changelog-generator), meant to replace this copy-pasted script once published; the script here carries a comment to swap it out when that lands.

Verified locally: ran the script against `v4.0.1..HEAD` (produced a correct entry, reverted before committing) and a full `npm run build` (repo.xml generates correctly and lands in the xar; the `.tmpl` is correctly excluded).

duncdrum:
Pushed the two follow-ups from the checklist:

- **Concurrency guard**: \`release.yml\` now serializes on a \`release\` group (\`cancel-in-progress: false\`), so two "exist-db CI" completions on develop landing close together queue instead of racing on version computation / the release commit / the tag push.
- **README**: added a "Commit message conventions" section explaining the Conventional Commits requirement (and that it's enforced via commitlint's PR check + the local husky hook, not just a suggestion), and rewrote "Publishing" — it still described the old manual tag-and-upload-to-public-repo process, which this PR replaces entirely.

line-o:
@duncdrum needs a rebase - other than that good to go in. :+1:

duncdrum:
@line-o rebased, ready to go once Ci is done

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "exide",
-	"version": "4.0.1",
+	"version": "0.0.0-development",
```
line-o: Monex does _not_ reset the version in package.json anymore.

line-o (approved):
I don't know if this should be blocking - but I honestly  would prefer we stay as close as possible to the workflow established in Monex

package.json
duncdrum: Done — restored `4.0.1` in `package.json` / lockfile. CI still bumps and commits it on release via `@semantic-release/git`, same as Monex.

PR #878, opened 2026-09-09
fix(view): xqdoc @ parse error and missing request import

duncdrum:
## Summary
- Reword the xqdoc comments in `modules/api/view.xqm` to avoid a raw `@` in prose (`restrictions/@guest`). eXist's ANTLR2 XQDoc lexer treats `@`+alphanumerics as a new tag, so `inspect:inspect-module` **silently truncates** the description mid-sentence (and a bare `@` can also yield `expecting XQDOC_END, found '@'`). Workaround for eXist-db/exist#1386; a short note in the comment guards against reverting the wording.
- Add the missing `import module namespace request="http://exist-db.org/xquery/request";` in `modules/api/view.xqm` — `view:render()` calls `request:get-context-path()` but the prefix was never declared. Many hosts predeclare `request` in `conf.xml` (so `XQST0081` may not always fire), but an explicit import is less fragile.

Found while testing #814 locally after merge.

## Test plan
- [x] `npm run build` succeeds and produces the XAR
- [x] Installed to a local eXist-db with `xst package install local-files --force`
- [x] Confirmed `exist.log` no longer shows the `XQDOC_END`/`@` parse error
- [x] Confirmed `inspect:inspect-module` returns the full `view:index` description (not truncated at the former `@`)
- [x] Confirmed no `XQST0081` for the `request` prefix where that error was observed


joewiz:

@duncdrum good catch, and thanks — this is my code from #814. I reproduced the `@` problem and it is worse than "a parse error in the log": the documentation is silently truncated.

Bed: `existdb/existdb:latest` (7.0.0-SNAPSHOT, build 2026-09-14T17:29Z), eXide `develop` at 4.1.0, packages as the image bundles them.

**On `develop`**, `inspect:inspect-module` returns `view:index`'s description cut off mid-sentence, exactly at the `@`:

```
Identity comes from `$request?user` (Roaster's standard-authorization). The
guest gate is config-driven (restrictions/
```

**With this branch**, the whole description comes back:

```
Identity comes from `$request?user` (Roaster's standard-authorization). The
guest gate is config-driven (the `restrictions` element's `guest` attribute),
so it can't be a static x-constraint; the handler checks auth:is-allowed
and, when the user is not allowed, redirects to login.html instead of
serving the app.
```

So the fix is doing real work — it just isn't only about a log message. The root cause is the ANTLR2 xqdoc lexer, which is your own eXist-db/exist#1386, open since 2017 and covering both the `@` truncation and the `%` failure a commenter added there. Since this is a workaround for an upstream bug rather than a style preference, it might be worth a short note in `view.xqm` pointing at eXist-db/exist#1386, so nobody later "tidies" the prose back to `restrictions/@guest` and silently loses the second half of the paragraph again.

Two smaller notes.

**`XQST0081` doesn't reproduce here.** eXist registers the request module in `conf.xml`, so the prefix resolves without an explicit import: `inspect:inspect-module` compiles the module and returns its five functions, and the app serves normally. The import is still right — leaning on a host-configured predeclaration is fragile — but it would be good to know what your instance does differently, since on mine the module is fine without it.

**There is a second, separate parser that fails outright.** `xqdoc:scan` (the bundled `org.xqdoc` library, not eXist's own `XQDocParser`) can't read any of eXide's API modules — it stops at the first XQuery 3.0/3.1 construct, before reaching any `@`:

| module | `xqdoc:scan` |
|---|---|
| `view.xqm` | `line 27:44: unexpected token: (` — `map(*)` |
| `auth.xqm` | `line 45:9: unexpected char: '%'` — `%private` |
| `db.xqm` | `line 25:79: unexpected token: {` — inline map |
| `admin.xqm` | `line 14:46: unexpected token: (` — `map(*)` |
| `editor.xqm` | `line 18:52: unexpected token: (` — `map(*)` |

That one is untouched by this PR (with the branch applied, `view.xqm` still fails at the same `map(*)`, just shifted to line 30), and it looks distinct from eXist-db/exist#1386, which is about eXist's own parser. If `xqdoc:scan` is a path anything actually depends on, that probably wants its own eXist issue — happy to file it with these numbers if you agree it isn't already covered.

For the rest: `view.xqm` is the only module in the repo with a raw `@` in xqdoc prose, so the reword is complete. Both commits look good to me.

joewiz:

Following up on my own open question — I offered to file an eXist issue for the `xqdoc:scan` failures if anything depended on that path. **Don't bother, and I withdraw the offer.** Digging in:

- `xqdoc:scan` comes from `extensions/xqdoc`, which depends on the external `org.xqdoc:xqdoc-conversion`. Maven Central has exactly **one release of it — 1.6, published 2012-08-17**. There is nothing newer to move to.
- Org-wide code search finds **no production caller** of `xqdoc:scan`. The only hit is `extensions/xqdoc/ant-example.xml` in eXist itself.

So it is dead weight rather than a live bug — a deprecation question for eXist at most, and nothing this PR should carry.

The path that *does* matter is the one your fix addresses, and I can now be precise about why it bites. Three rules in `XQDocParser.g` (ANTLR 2.4, 1.9 KB):

```antlr
CHARS: (~('\n' | '@' | ':'))+;     // prose cannot contain '@' at all
TAG:   '@' ('A'..'Z' | 'a'..'z' | '0'..'9')+;
AT:    '@';                         // declared, never referenced by any parser rule
```

`CHARS` excludes `@`, so what happens next depends on the following character. `@` plus alphanumerics — `@guest`, `@type` — lexes as `TAG`, and since `xqdocComment` has `( TAG ) => taggedContents[doc]`, the parser treats it as the start of a *new xqdoc tag*: the rest of the paragraph goes to `doc.setTag("@guest", …)` instead of `addDescription`, which is the silent truncation I measured. A `@` followed by anything else emits `AT`, which no parser rule accepts — that looks like the origin of the `expecting XQDOC_END, found '@'` you hit, so the two of us were seeing the same defect through its two exits.

Minimal repro, no eXide involved:

```xquery
(:~
 : Selects taxonomy[@type = "reign"] from the source.
 : THIS SENTENCE MUST SURVIVE.
 :)
```

`inspect:inspect-module` returns `Selects taxonomy[` — and a bare `@` truncates too (`Costs 5 @ 3 dollars each. SECOND SENTENCE.` gives `Costs 5`).

One more thing for eXist-db/exist#1386, since it's yours: the `%` half looks already resolved. `inspect:inspect-module` handles `%private`-annotated modules without error on current `develop` — all five of eXide's API modules inspect cleanly — so the issue could probably be narrowed to the `@` case. Joe has a root-cause write-up going to the core side; nothing for this PR to wait on.

None of this changes my read of the PR — both commits still look good, and the description trim is the only suggestion I'd keep.

duncdrum:
@joewiz done, we can continue in the exist issue as far as fixing the underlying bug is concerned.

PR #880, opened 2026-09-14
ci(release): use client-id for create-github-app-token

duncdrum:
## Summary
- Switch `actions/create-github-app-token@v3` from deprecated `app-id` to `client-id` to clear the CI warning.
- Uses the existing org variable `vars.RELEASE_BOT_CLIENT_ID`. No App or private-key changes.

## Test plan
- [ ] After merge to `develop`, next successful `exist-db CI` run triggers Semantic Release without the `app-id` deprecation warning
- [ ] Token minting / release push still works if a release is due


duncdrum:
@line-o here it is, the client-id is not an org variable like the app-id. We should probably merge this after the token got permissions to comment on PR and Issues.

line-o (approved):

